PR #1, opened 2022-07-06
Add a Dockerfile for building a numba_dpex dev environment

fcharras:
This Dockerfile should enable a working environment with a set of nice properties:
- a standard environment
- easy installation
- reproducibility
- configurability (set python, dpctl, dpnp, numba-dpex, intel numpy, intel scipy, intel numba versions)
and it gives an overview of the packaging of intel softwares used for numba-dpex and their contents.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,21 +35,21 @@ ARG BASE=${BASE_IMAGE_TAG}:${BASE_IMAGE_VERSION}
 ARG INTEL_PYPI_URL=https://pypi.anaconda.org/intel/simple
 ARG BASE_PYPI_URL=https://pypi.org/simple
 
-# /!\ the following version must exist in intel pypi. It is recommended to
+# /!\ the following versions must exist in intel pypi. It is recommended to
 # not use version ranges, because if a higher version exists in pypi.org it
 # will be preferred over the versions in the intel pypi.
 
 ARG INTEL_NUMPY_VERSION="==1.22.3"
+ARG INTEL_SCIPY_VERSION="==1.7.3"
 
 # XXX: This can be fixed if intel can host the whole the dependency tree
 # in the INTEL_PYPI_URL instead of having to rely in pypi.org for the main part
 # of the dependency tree.
 
-# ???: it is not clear so far if those versions should also be limited to the 
+# ???: it is not clear so far if this version should also be limited to the 
 # content of the intel pypi or not. (this Dockerfile currently assumes that
 # it is not limited to intel pypi. The opposite assumption seems to create 
 # impossible build conditions.)
-ARG INTEL_SCIPY_VERSION=""
 ARG INTEL_NUMBA_VERSION="<0.56"
 
 
@@ -108,7 +108,7 @@ ARG ONEAPI_LOG_DIR=$TMPDIR/intel/log
 ARG ONEAPI_COMPONENTS="intel.oneapi.lin.dpcpp-cpp-compiler"
 RUN mkdir -p $ONEAPI_INSTALL_DIR $ONEAPI_CACHE_DIR $ONEAPI_DOWNLOAD_DIR ONEAPI_LOG_DIR \
     && wget -P $ONEAPI_DOWNLOAD_DIR $ONEAPI_INSTALLER_URL/$ONEAPI_INSTALL_BINARY_NAME \
-    && chmod +x $ONEAPI_DOWNLOAD_DIR/l_BaseKit_p_2022.2.0.262.sh \
+    && chmod +x $ONEAPI_DOWNLOAD_DIR/$ONEAPI_INSTALL_BINARY_NAME \
     && $ONEAPI_DOWNLOAD_DIR/$ONEAPI_INSTALL_BINARY_NAME -a -s --eula accept \
         --action install --components $ONEAPI_COMPONENTS  \
         --install-dir $ONEAPI_INSTALL_DIR --log-dir $ONEAPI_LOG_DIR \
@@ -127,7 +127,7 @@ ARG ONEAPI_LOG_DIR=$TMPDIR/intel/log
 ARG ONEAPI_COMPONENTS="intel.oneapi.lin.dpcpp-cpp-compiler:intel.oneapi.lin.tbb.devel:intel.oneapi.lin.mkl.devel"
 RUN mkdir -p $ONEAPI_INSTALL_DIR $ONEAPI_CACHE_DIR $ONEAPI_DOWNLOAD_DIR ONEAPI_LOG_DIR \
     && wget -P $ONEAPI_DOWNLOAD_DIR $ONEAPI_INSTALLER_URL/$ONEAPI_INSTALL_BINARY_NAME \
-    && chmod +x $ONEAPI_DOWNLOAD_DIR/l_BaseKit_p_2022.2.0.262.sh \
+    && chmod +x $ONEAPI_DOWNLOAD_DIR/$ONEAPI_INSTALL_BINARY_NAME \
     && $ONEAPI_DOWNLOAD_DIR/$ONEAPI_INSTALL_BINARY_NAME -a -s --eula accept \
         --action modify --components $ONEAPI_COMPONENTS  \
         --install-dir $ONEAPI_INSTALL_DIR --log-dir $ONEAPI_LOG_DIR \
@@ -232,7 +232,8 @@ COPY --from=python_builder $VIRTUAL_ENV $VIRTUAL_ENV
 # pypi.
 # The proper fix is for intel to host all the dependency tree.
 # see https://stackoverflow.com/questions/67253141/python-pip-priority-order-with-index-url-and-extra-index-url
-RUN pip install -U -i $INTEL_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" --no-deps \
+RUN pip install -U -i $INTEL_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" \
+        "scipy${INTEL_SCIPY_VERSION}" --no-deps \
     && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL \
         "numpy${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" \
         "scipy${INTEL_SCIPY_VERSION}" \ 
@@ -333,7 +334,8 @@ ARG PACKAGING_VERSION
 COPY --from=dpctl_builder $TMPDIR/dpctl*.whl $TMPDIR
 COPY --from=dpnp_builder $TMPDIR/dpnp*.whl $TMPDIR
 COPY --from=numba_dpex_builder $TMPDIR/numba_dppy*.whl $TMPDIR
-RUN pip install -U -i $INTEL_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" --no-deps \
+RUN pip install -U -i $INTEL_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" \
+        "scipy${INTEL_SCIPY_VERSION}" --no-deps \
     && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL \
         "numpy${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" \
         "scipy${INTEL_SCIPY_VERSION}" $TMPDIR/dpctl*.whl $TMPDIR/dpnp*.whl \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,22 +1,28 @@
-# This Dockerfile can be built from scratch with e.g the command
-# docker build . -t my_numba_dpex_build
+# This Dockerfile can be built from scratch with e.g. the command
+#
+#     docker build . -t numba_dpex_dev
+#
 # The ARGs in the header of this file can be changed as needed (either by
 # editing the file or with the command line).
 # The build can be fairly long (up to one hour on low-end machines.)
-# To start a container, use e.g
-# sudo docker run --rm --name my_container -it -v /my/host/volume/:/mounted/volume --device=/dev/dri my_numba_dpex_build
-# The parameter --device=/dev/dri enables gpu passthrough (as long as the user 
-# has permissions, e.g by being a member of the video group)
+# To start a container, use e.g.:
+# 
+#      sudo docker run --rm --name my_container -it -v /my/host/volume/:/mounted/volume --device=/dev/dri numba_dpex_dev
+#
+# The --device=/dev/dri option enables GPU passthrough (as long as the user 
+# has permissions, e.g. by being a member of the video group)
 
-
-# TODO: make a custom build of the base image to test with other base OS
-# (e.g ubuntu2204)
-# ???: actually not sure if it's needed or interesting to use the sycl nightly
-# binaries at all. The nightly also ship an up-to-date installation of the
+# The `ubuntu2004_intel_drivers` ships an up-to-date installation of the
 # intel runtime, including drivers and opencl and level_zero runtimes.
+#
+# See available versions here: https://github.com/intel/llvm/pkgs/container/llvm%2Fubuntu2004_intel_drivers
+#
+# TODO: make a custom build with other base images to test compatibiliry
+# with other OSes
+ARG BASE_IMAGE_VERSION=latest-72882d570b860258739a83bc23609c6f6dd3002f
+ARG BASE_IMAGE_TAG=ghcr.io/intel/llvm/ubuntu2004_intel_drivers
 
-ARG BASE=ghcr.io/intel/llvm/sycl_ubuntu2004_nightly
-ARG TMPDIR=/tmp
+ARG BASE=${BASE_IMAGE_TAG}:${BASE_IMAGE_VERSION}
 
 # Various resources about the SYCL toolchain
 # Environment variables with important effects on the SYCL toolchain:
@@ -35,15 +41,14 @@ ARG BASE_PYPI_URL=https://pypi.org/simple
 
 ARG INTEL_NUMPY_VERSION="==1.22.3"
 
-# TODO: This can be fixed if intel can host the whole the dependency tree
+# XXX: This can be fixed if intel can host the whole the dependency tree
 # in the INTEL_PYPI_URL instead of having to rely in pypi.org for the main part
 # of the dependency tree.
-# TODO: find which part of the intel basekit should be installed
-# with the installer, and which part can be used from python packages, at
-# build time and at run time, might enable a lighter build.
 
-# ???: not clear if those versions should also be limited to the content
-# of the intel pypi or not. (looks like not)
+# ???: it is not clear so far if those versions should also be limited to the 
+# content of the intel pypi or not. (this Dockerfile currently assumes that
+# it is not limited to intel pypi. The opposite assumption seems to create 
+# impossible build conditions.)
 ARG INTEL_SCIPY_VERSION=""
 ARG INTEL_NUMBA_VERSION="<0.56"
 
@@ -56,9 +61,10 @@ ARG PYTHON_VERSION=3.9.13
 # It can be found at https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html?operatingsystem=linux&distributions=webdownload&options=online
 # The installer CLI is documented here: https://www.intel.com/content/www/us/en/develop/documentation/installation-guide-for-intel-oneapi-toolkits-linux/top/installation/install-with-command-line.html#install-with-command-line_interactive
 
-ARG ONEAPI_INSTALLER=https://registrationcenter-download.intel.com/akdlm/IRC_NAS/18673/l_BaseKit_p_2022.2.0.262.sh
-ARG ONEAPI_INSTALL_DIR=/opt/intel/oneapi
+ARG ONEAPI_INSTALLER_URL=https://registrationcenter-download.intel.com/akdlm/IRC_NAS/18673
+ARG ONEAPI_INSTALL_BINARY_NAME=l_BaseKit_p_2022.2.0.262.sh
 
+ARG ONEAPI_INSTALL_DIR=/opt/intel/oneapi
 
 # Some build steps requires a high enough version of cmake.
 # Bump it if necessary.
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,291 @@
...
+ARG INTEL_NUMPY_VERSION
+ARG INTEL_SCIPY_VERSION
+ARG INTEL_NUMBA_VERSION
+COPY --from=dpctl_builder $TMPDIR/dpctl*.whl $TMPDIR/
+COPY --from=dpnp_builder $TMPDIR/dpnp*.whl $TMPDIR/
+COPY --from=numba_dpex_builder $TMPDIR/numba_dpex*.whl $TMPDIR/
+RUN apt-get update --quiet && apt-get install -y --no-install-recommends spirv-tools spirv-headers \
+    && rm -Rf /var/lib/apt/lists/* \
+    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy==${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" $TMPDIR/dpctl*.whl $TMPDIR/dpnp*.whl $TMPDIR/numba_dpex*.whl cython packaging pytest pexpect \
+    && rm -Rf $TMPDIR/*
```
ogrisel, line 284: I tried to build without passing any args (`docker build -t dpex .`) and got the following error:

```
WARNING: The index url "numpy====1.22.3" seems invalid, please provide a scheme.
Looking in indexes: https://pypi.anaconda.org/intel/simple, numpy====1.22.3
WARNING: Location 'numpy====1.22.3/numba/' is ignored: it is either a non-existing path or lacks a specific scheme.
Collecting numba
  Downloading https://pypi.anaconda.org/intel/simple/numba/0.54.1/numba-0.54.1-4-cp39-cp39-manylinux2014_x86_64.whl (2.3 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.3/2.3 MB 1.5 MB/s eta 0:00:00
WARNING: Location 'numpy====1.22.3/scipy/' is ignored: it is either a non-existing path or lacks a specific scheme.
Collecting scipy
  Downloading https://pypi.anaconda.org/intel/simple/scipy/1.7.3/scipy-1.7.3-0-cp39-cp39-manylinux_2_17_x86_64.whl (31.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 31.8/31.8 MB 879.5 kB/s eta 0:00:00
WARNING: Location 'numpy====1.22.3/wheel/' is ignored: it is either a non-existing path or lacks a specific scheme.
ERROR: Could not find a version that satisfies the requirement wheel (from versions: none)
ERROR: No matching distribution found for wheel
WARNING: Location 'numpy====1.22.3/pip/' is ignored: it is either a non-existing path or lacks a specific scheme.
WARNING: There was an error checking the latest version of pip.
The command '/bin/sh -c apt-get update --quiet     && apt-get install -y --no-install-recommends         build-essential         libtool         autoconf         unzip         libssl-dev         wget         ninja-build     && rm -rf /var/lib/apt/lists/*     && apt purge -y --auto-remove cmake     && mkdir -p $CMAKE_BUILD_DIR     && cd $CMAKE_BUILD_DIR     && wget https://cmake.org/files/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}.tar.gz     && tar -xzvf cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}.tar.gz     && cd cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}     && ./bootstrap     && make -j$(nproc)     && sudo make install     && rm -Rf $CMAKE_BUILD_DIR     && cd     && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy==${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" wheel cython scikit-build' returned a non-zero code: 1
```

docker/Dockerfile
```diff
@@ -0,0 +1,291 @@
...
+
+
+ARG INTEL_PYPI_URL=https://pypi.anaconda.org/intel/simple
+ARG BASE_PYPI_URL=https://pypi.org/simple
+
+# /!\ the following version must exist in intel pypi. It is recommended to
+# not use version ranges, because if a higher version exists in pypi.org it
+# will be preferred over the versions in the intel pypi.
+
+ARG INTEL_NUMPY_VERSION="==1.22.3"
```
jjerphan, line 36: This should resolve [the problem mentioned by @ogrisel](https://github.com/ogrisel/sklearn-numba-dpex/pull/1/files#r914945843)
```suggestion
ARG INTEL_NUMPY_VERSION="1.22.3"
```

docker/Dockerfile
```diff
@@ -0,0 +1,291 @@
...
+    && cd $CMAKE_BUILD_DIR \
+    && wget https://cmake.org/files/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}.tar.gz \
+    && tar -xzvf cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}.tar.gz \
+    && cd cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD} \
+    && ./bootstrap \
+    && make -j$(nproc) \
+    && sudo make install \
+    && rm -Rf $CMAKE_BUILD_DIR \
+    && cd \
+    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy==${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" wheel cython scikit-build
```
jjerphan, line 188: This should resolve [the problem mentioned by @ogrisel](https://github.com/ogrisel/sklearn-numba-dpex/pull/1/files#r914945843)

```suggestion
    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" wheel cython scikit-build
```

docker/Dockerfile
```diff
@@ -0,0 +1,291 @@
...
+ARG BASE_PYPI_URL
+ARG INTEL_NUMPY_VERSION
+ARG INTEL_SCIPY_VERSION
+ARG INTEL_NUMBA_VERSION
+COPY --from=dpctl_builder $TMPDIR/dpctl*.whl $TMPDIR/
+COPY --from=dpnp_builder $TMPDIR/dpnp*.whl $TMPDIR/
+COPY --from=numba_dpex_builder $TMPDIR/numba_dpex*.whl $TMPDIR/
+RUN apt-get update --quiet && apt-get install -y --no-install-recommends spirv-tools spirv-headers \
+    && rm -Rf /var/lib/apt/lists/* \
+    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy==${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" $TMPDIR/dpctl*.whl $TMPDIR/dpnp*.whl $TMPDIR/numba_dpex*.whl cython packaging pytest pexpect \
```
jjerphan, line 283: Similarly.
```suggestion
    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" $TMPDIR/dpctl*.whl $TMPDIR/dpnp*.whl $TMPDIR/numba_dpex*.whl cython packaging pytest pexpect \
```

fcharras:
Looks like I've let a few mistakes slip through the pushed version. I'm fixing it and will push it when the build is successful on my end. Thank you for the reports.

fcharras:
Latest version pushed builds successfully on my end could you try again ?

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+ARG CMAKE_VERSION=3.23
+ARG CMAKE_VERSION_BUILD=2
+
+
+# Versions of the intel python packages
+
+ARG DPCTL_GIT_BRANCH=0.13.0dev2
+ARG DPCTL_GIT_URL=https://github.com/IntelPython/dpctl.git
+
+ARG DPNP_GIT_BRANCH=master
```
jjerphan, line 75: The `master` branch potentially might change between image builds.
Must we use tags or even commits instead to reference specific versions and hence reproducible builds?

If so, we might want to enforce this for other projects' branches as well.

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+
+# So far, the intel numpy python package is not built for python>=3.10
+ARG PYTHON_VERSION=3.9.13
+
+
+# This link might need to be refreshed occasionally.
+# It can be found at https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html?operatingsystem=linux&distributions=webdownload&options=online
+# The installer CLI is documented here: https://www.intel.com/content/www/us/en/develop/documentation/installation-guide-for-intel-oneapi-toolkits-linux/top/installation/install-with-command-line.html#install-with-command-line_interactive
+
+ARG ONEAPI_INSTALLER=https://registrationcenter-download.intel.com/akdlm/IRC_NAS/18673/l_BaseKit_p_2022.2.0.262.sh
```
jjerphan, line 59: Since the base kit script is reused, is is possible to define it like this?

```suggestion
ARG BASEKIT_SCRIPT=l_BaseKit_p_2022.2.0.262.sh
ARG ONEAPI_INSTALLER=https://registrationcenter-download.intel.com/akdlm/IRC_NAS/18673/$BASEKIT_SCRIPT
```

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+ARG ONEAPI_DOWNLOAD_DIR=$TMPDIR/intel/download
+ARG ONEAPI_LOG_DIR=$TMPDIR/intel/log
+# TODO: are all those components necessary ?
+# TODO: some of those components that are necessary for build might not be
+# necessary for runtime ?
+ARG ONEAPI_COMPONENTS="intel.oneapi.lin.dpcpp-ct:intel.oneapi.lin.dpcpp_dbg:intel.oneapi.lin.dpcpp-cpp-compiler:intel.oneapi.lin.tbb.devel:intel.oneapi.lin.mkl.devel"
+RUN mkdir -p $ONEAPI_INSTALL_DIR $ONEAPI_CACHE_DIR $ONEAPI_DOWNLOAD_DIR ONEAPI_LOG_DIR \
+    && wget -P $ONEAPI_DOWNLOAD_DIR $ONEAPI_INSTALLER \
+    && chmod +x $ONEAPI_DOWNLOAD_DIR/l_BaseKit_p_2022.2.0.262.sh \
+    && $ONEAPI_DOWNLOAD_DIR/l_BaseKit_p_2022.2.0.262.sh -a -s --eula accept --action install --components $ONEAPI_COMPONENTS --install-dir $ONEAPI_INSTALL_DIR --log-dir $ONEAPI_INSTALL_DIR --download-cache $ONEAPI_CACHE_DIR \
```
jjerphan, line 96:
```suggestion
ARG TMPDIR
ARG ONEAPI_INSTALL_DIR
ARG ONEAPI_INSTALLER
ARG BASEKIT_SCRIPT
ARG ONEAPI_CACHE_DIR=$TMPDIR/intel/cache
ARG ONEAPI_DOWNLOAD_DIR=$TMPDIR/intel/download
ARG ONEAPI_LOG_DIR=$TMPDIR/intel/log
# TODO: are all those components necessary ?
# TODO: some of those components that are necessary for build might not be
# necessary for runtime ?
ARG ONEAPI_COMPONENTS="intel.oneapi.lin.dpcpp-ct:intel.oneapi.lin.dpcpp_dbg:intel.oneapi.lin.dpcpp-cpp-compiler:intel.oneapi.lin.tbb.devel:intel.oneapi.lin.mkl.devel"
RUN mkdir -p $ONEAPI_INSTALL_DIR $ONEAPI_CACHE_DIR $ONEAPI_DOWNLOAD_DIR ONEAPI_LOG_DIR \
    && wget -P $ONEAPI_DOWNLOAD_DIR $ONEAPI_INSTALLER \
    && chmod +x $ONEAPI_DOWNLOAD_DIR/BASEKIT_SCRIPT \
    && $ONEAPI_DOWNLOAD_DIR/BASEKIT_SCRIPT -a -s --eula accept --action install --components $ONEAPI_COMPONENTS --install-dir $ONEAPI_INSTALL_DIR --log-dir $ONEAPI_INSTALL_DIR --download-cache $ONEAPI_CACHE_DIR \
```

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+        tk-dev \
+        libxml2-dev \
+        libxmlsec1-dev \
+        libffi-dev \
+        liblzma-dev \
+        git \
+        ca-certificates \
+     && rm -rf /var/lib/apt/lists/*
+RUN curl -L \
+    https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer \
```
jjerphan, line 135: Should we get this file from a specific tag or commit? e.g:
```suggestion
    https://github.com/pyenv/pyenv-installer/raw/49fba599e872bc761858ea6f700271fb6dcb5a97/bin/pyenv-installer \
```

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+    && cd $CMAKE_BUILD_DIR \
+    && wget https://cmake.org/files/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}.tar.gz \
+    && tar -xzvf cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD}.tar.gz \
+    && cd cmake-${CMAKE_VERSION}.${CMAKE_VERSION_BUILD} \
+    && ./bootstrap \
+    && make -j$(nproc) \
+    && sudo make install \
+    && rm -Rf $CMAKE_BUILD_DIR \
+    && cd \
+    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" wheel cython scikit-build
```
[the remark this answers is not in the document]
jjerphan: Also, should we pin `wheel`, `cython` and `scikit-build`?

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+ARG BASE_PYPI_URL
+ARG INTEL_NUMPY_VERSION
+ARG INTEL_SCIPY_VERSION
+ARG INTEL_NUMBA_VERSION
+COPY --from=dpctl_builder $TMPDIR/dpctl*.whl $TMPDIR
+COPY --from=dpnp_builder $TMPDIR/dpnp*.whl $TMPDIR
+COPY --from=numba_dpex_builder $TMPDIR/numba_dppy*.whl $TMPDIR
+RUN apt-get update --quiet && apt-get install -y --no-install-recommends spirv-tools spirv-headers \
+    && rm -Rf /var/lib/apt/lists/* \
+    && pip install -U -i $INTEL_PYPI_URL --extra-index-url $BASE_PYPI_URL "numpy${INTEL_NUMPY_VERSION}" "numba${INTEL_NUMBA_VERSION}" "scipy${INTEL_SCIPY_VERSION}" $TMPDIR/dpctl*.whl $TMPDIR/dpnp*.whl $TMPDIR/numba_dppy*.whl cython packaging pytest pexpect \
```
jjerphan, line 285: Similarly, should we pin `cython`, `packaging`, `pytest` and `pexpect`?

docker/Dockerfile
jjerphan, line 96:
```suggestion
    && $ONEAPI_DOWNLOAD_DIR/l_BaseKit_p_2022.2.0.262.sh -a -s --eula accept --action install --components $ONEAPI_COMPONENTS --install-dir $ONEAPI_INSTALL_DIR --log-dir $ONEAPI_LOG_DIR --download-cache $ONEAPI_CACHE_DIR \
```

jjerphan (commented):
Thank you, @fcharras.

It successfully builds on my end! The image is a bit large at the moment, but I guess we can trim it.

```
$ sudo docker images
REPOSITORY                                   TAG       IMAGE ID       CREATED          SIZE
my_numba_dpex_build                          latest    580ff85a28bf   7 minutes ago    14.3GB
```

Here are a few comments and questions.

---

To come back to your questions and remarks:
> One point to be careful about is that with the 0.18.0 tag (which is the most recent tag) numba_dpex builds as numba_dppy and not numba_dpex, is numba_dpex the most recent name ?

numba_dpex previous name was numba_dppy. The renaming was done, but the package is not yet distributed with the latest name when we last checked.

> is the package intended to be built from master rather than 0.18.0 ? if you replace "0.18.0" by "master" it's probably needed to also rewrite "numba_dppy" as "numba_dpex" at the end of the file. Likewise, the examples from master will not work with 0.18.0 because those examples import numba_dpex and not numba_dppy (but the examples from the 0.18.0 checkout do work).

I think it's fine to stick to `0.18.0` first, then we can try to work using the latest version.

docker/Dockerfile
```diff
@@ -0,0 +1,291 @@
...
+
+
+# TODO: make a custom build of the base image to test with other base OS
+# (e.g ubuntu2204)
+# ???: actually not sure if it's needed or interesting to use the sycl nightly
+# binaries at all. The nightly also ship an up-to-date installation of the
+# intel runtime, including drivers and opencl and level_zero runtimes.
+
+ARG BASE=ghcr.io/intel/llvm/sycl_ubuntu2004_nightly
+ARG TMPDIR=/tmp
```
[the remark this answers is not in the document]
fcharras: Thank you for rewording, a lot of comments are still personal notes I wrote down in the process rather than appropriate documentation, and the file is not in a good shape overall.

About the use of sycl_ubuntu2004_nightly, to begin with I chose the image because it ships up-to-date drivers and runtime for opencl and level_zero, and for now it strips away the complexity of those steps from the current PR, which can be more focused on getting the numba-dpex installation right. The dockerfiles for this base are included in the llvm repo and looks sensible which made me think this base is reliable + it benefits from daily CI. It's of course possible to start from any other base (preferably debian-like) image and script the driver installation (it's will even be necessary to test compatibility with other OSes), at the cost of some more complexity.

If the use of the nightly sycl binaries turns out to be unneeded or detrimental, `ghcr.io/intel/llvm/ubuntu2004_intel_drivers` can be used instead, it's the same base without sycl binaries.

You're right about pinning the version of the base image, will add.

docker/Dockerfile
```diff
@@ -0,0 +1,294 @@
...
+ARG DPCTL_BUILD_DIR=$TMPDIR/dpctl
+ARG SKBUILD_ARGS="-- -DCMAKE_C_COMPILER:PATH=icx -DCMAKE_CXX_COMPILER:PATH=icpx"
+RUN mkdir -p $DPCTL_BUILD_DIR \
+    && cd $DPCTL_BUILD_DIR \
+    && git clone --recursive -b $DPCTL_GIT_BRANCH --depth 1 $DPCTL_GIT_URL . \
+    && source $ONEAPI_INSTALL_DIR/setvars.sh \
+    && export PATH="/opt/sycl/bin:$VIRTUAL_ENV/bin:$PATH" \
+    && export LD_LIBRARY_PATH="/opt/sycl/lib:$VIRTUAL_ENV/lib:$LD_LIBRARY_PATH" \
+    && python setup.py clean --all \
+    # TODO: manylinux wheel build arg ?
```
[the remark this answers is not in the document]
fcharras: There are several specifications for python package builds, like [manylinux](https://github.com/pypa/manylinux), and intel conda builds [explicitly name them ](https://github.com/IntelPython/dpnp/blob/master/conda-recipe/build.sh#L35).

I'm not familiar with what it means and essentially skipped the step, and let this `TODO` as a reminder to look at it later - will clarify the TODO.

docker/Dockerfile
fcharras: > About pinning versions for wheel, cython and scikit-build, while it doesn't cost much to expose ARGs for those (will do), I thought those packages are fairly stables and shouldn't be a concern anyway. Conda recipes on intel side do not bother with pinning those either.
Likewise for those packages

fcharras:
And about the size of the image, there surely is place for trimming, I think some binaries are installed two or even three times...
- one time within a oneAPI basekit component
- one time within a python package (intel numpy ships a lot of binaries such as mkl and dpcpp runtimes as python packages hosted on pypi.org)
- one time within the SYCL nightly binaries
when we find out which binaries are necessary and can uninstall the others the image should significantly decrease in size.

fcharras:
I've pushed several improvements:
- overall cleaning and iterations following @jjerphan comments (thank you for the review)
- changed the base image to ghcr.io/intel/llvm/ubuntu2004_intel_drivers and removed the sycl nightly binary
- filtered the intel basekit components that are not needed
- removed the big copy step that triggered swapping at the end of the build (reported by @ogrisel)
- fixed a bug where numpy was installed from pypi.org rather than intel pypi which also caused a lot of missing python dependencies

The image is 3GB lighter, there's probably about 3GB more to remove because the dpcpp compiler is still installed with the basekit but only the llvm-spirv binary is really being used. 

Maybe we could merge this version and host the image in a registry ?

jjerphan (approved):
Thank you, @fcharras!

I would be pragmatic: the overall setup is complex but the image works: LGTM — I would not have done better!

IMO, this can further be refined with discussions with Intel's engineers — potentially there's a simpler setup?

docker/Dockerfile
```diff
@@ -0,0 +1,350 @@
...
+        "scipy${INTEL_SCIPY_VERSION}" $TMPDIR/dpctl*.whl $TMPDIR/dpnp*.whl \
+        $TMPDIR/numba_dppy*.whl packaging${PACKAGING_VERSION} \
+    && pip cache purge
+
+FROM runtime_environment AS numba_dpex_dev
+COPY --from=numba_dpex_dev_installer $VIRTUAL_ENV $VIRTUAL_ENV
+RUN apt-get update --quiet && apt-get install -y --no-install-recommends \
+        spirv-tools spirv-headers \
+    && rm -Rf /var/lib/apt/lists/*
+CMD ["/bin/bash"]
```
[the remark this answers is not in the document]
fcharras: We don't need it because the libraries are shipped within python packages as dependencies of intel numpy (mkl, dpcpp_rt, etc), and the path to the lib folder in the venv is added to `LD_LIBRARY_PATH` already.

The only part of the runtime that are not shipped with python packages and that are taken from the oneapi basekit (in fact, only the compiler component of the basekit) are the executables in the folder bin-llvm, that are added to the $PATH. This is hinted [here](https://github.com/IntelPython/numba-dpex#known-issue) and [here](https://github.com/IntelPython/numba-dpex/blob/main/conda-recipe/build.sh#L7).

This is the way I understand the build, but I'd advise to be cautious and to check that the jit in `numba_dpex` works as expected when using it (e.g I hope there are no such things as a silent fallback mechanism if some libraries are missing) until there's enough coverage to certify that it's the right build :)

fcharras:
> IMO, this can further be refined with discussions with Intel's engineers — potentially there's a simpler setup?

Agree. On intel's end the simpler setup is to use conda, but I think a conda install should be an option to the user and not the only way. Though we can probably make a docker build work by using conda within the container.

This build is complicated but it gives the overview over the underlying dependency tree since it builds with basic tools and packaging (make & python setup.py), I find it to be a more comfortable developing environment because if something goes wrong I can refer directly to the basic installation steps for debugging.

Another way to make it simpler is to not have to build from source packages such as `dpnp`, `dpctl` and `numba_dpex` or even have to install the `intel basekit` (as long as a development environment for one of those packages is not needed), and have everything `pip install`-able, but currently it looks like an exhaustive package index with up-to-date binaries for all python versions is missing. (see (dpnp on pypi)[https://pypi.org/project/dpnp/] which as not been updated for a while)

I have a few pending questions and remarks:

- is there any differences between the numba package hosted on intel pypi and normal numba ? ([this](https://pypi.anaconda.org/intel/simple/numba/) and [this](https://github.com/IntelPython/numba-dpex/blob/main/conda-recipe/meta.yaml#L21) hints that for version (`>=55`) the package from the official pypi should be used.

- how are `llvmlite` (python package), `llvmdev` (conda forge feedstock) and binaries within the compiler component from the oneapi basekit related to each other, is it fine to not use `llvmdev` but only the executables within the `bin-llvm` folder of the latest oneapi basekit version ?

- the installation could be made easier if intel hosts all of the python packages within the dependency tree of their packaging of scipy and numpy, because pip does not supports resolving dependency trees where packages with the same name appears in different package indexes. (+ supporting a wider range of python versions all the versions)

- rather than having to add `$VIRTUAL_ENV/lib` to `$LD_LIBRARY_PATH` shouldn't the python packages automatically enforce it in the python runtime ?

fcharras:
Changed:
- removed --with-computed-gotos when installing python
- added missing mkl-dpcpp python package

ogrisel:
I think it's a good first iteration. I pulled the image and it works for me (that is, I can import dpct, dpnp, numba_dppy and reproduce the same error I reported under https://github.com/IntelPython/numba-dpex/issues/756 from within this docker container).

ogrisel (approved):
Let's merge as is. We can always refine later. In particular by not building cmake from source for instance.

PR #4, opened 2022-07-28
CI Add minimal GitHub Actions workflow to run tests

jjerphan:
This add a minimal CI setup using a GitHub action workflow to run tests.

[The run completes](https://github.com/jjerphan/sklearn-numba-dpex/actions/runs/2816525277) with this simple setup.

It uses one of [intel's images for oneapi](https://hub.docker.com/r/intel/oneapi-basekit/tags), namely:

```
intel/oneapi-basekit:2022.2-devel-ubuntu20.04
```


sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -6,9 +8,14 @@
 from sklearn.utils._testing import assert_allclose
 
 
+def test_placeholder():
+    """Placeholder test for CI"""
```
fcharras, line 12: maybe the following first test that checks that the main packages are installed:

```
try:
    import numba_dpex as dpex
except ImportError:
    import numba_dppy as dpex

import dpctl

dpctl.get_devices()

```
?

.github/workflows/run_tests.yaml
```diff
@@ -0,0 +1,22 @@
+name: Run tests
+on:
+  pull_request:
+  push: { branches: main }
+
+jobs:
+  test:
+    name: Run test suite
+    runs-on: ubuntu-latest
+    container: intel/oneapi-basekit:2022.2-devel-ubuntu20.04
```
fcharras, line 10: if size is not an issue we could use our own image after all ?

fcharras (approved):

sklearn_numba_dpex/tests/test_kmeans.py
jjerphan: Yes, let's add those imports for this test. :+1:

On a side-note, you should be able to use GitHub suggestions. For instance:

```suggestion
    """Placeholder test for CI"""
    try:
	    import numba_dpex as dpex
	except ImportError:
	    import numba_dppy as dpex
	
	import dpctl
	
	# There must be at least one usable device.
	assert len(dpctl.get_devices()) > 0
```

.github/workflows/run_tests.yaml
jjerphan: Yes, I've found the issue I add.
I will push another commit soon.

ogrisel (approved):
I am fine merging this as is to get started and improve later to use our own docker image. As you prefer.

ogrisel:
The merge commit has triggered a run here:

https://github.com/soda-inria/sklearn-numba-dpex/runs/7745901558?check_suite_focus=true

PR #8, opened 2022-08-11
updating Dockerfile to latest dpctl, dpnp and numba_dpex releases

fcharras:
The 0.18.1 version from "numba_dpex" switches from `numba_dppy` module name to `numba_dpex`, from then on imports must refer to the latter.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -75,13 +75,13 @@ ARG CMAKE_VERSION_BUILD=2
 
 # Versions of the intel python packages
 
-ARG DPCTL_GIT_BRANCH=0.13.0dev2
+ARG DPCTL_GIT_BRANCH=0.13.0
 ARG DPCTL_GIT_URL=https://github.com/IntelPython/dpctl.git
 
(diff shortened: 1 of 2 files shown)


jjerphan (approved):
LTGM.

I think the imports in the placeholder test can also be changed in this PR to only import `numba_dpex`.

fcharras:
- [x] Changed the placeholder test

fcharras:
With this new change, the test will not pass unless we push the updated image to dockerhub and use it here

ogrisel (approved):

ogrisel:
I assume the following will be fixed when uploading the new version of the image on the gitlab registry.

```
_______________________________ test_placeholder _______________________________

    def test_placeholder():
        """Placeholder test for CI"""
>       import numba_dpex as dpex
E       ModuleNotFoundError: No module named 'numba_dpex'
```

fcharras:
yes (see comment above) in fact we wanted to update the image on dockerhub and created a group but I don't have access yet. Will update the image as soon as it's available on dockerhub.

PR #10, opened 2022-08-18
Use F-layout for X (transposed) in device memory

ogrisel:
@fcharras I think you intended to use the Fortran layout for the transposed X but you did not.

I tried and on my machine this results in a small performance improvement when running the benchmark.

### DIFF
--- a/sklearn_numba_dpex/kmeans/drivers.py
+++ b/sklearn_numba_dpex/kmeans/drivers.py
@@ -156,8 +156,8 @@ def kmeans(
         number_of_load_iter=4,
     )
 
-    # The dpex kernel expects X to be shaped as (n_features, n_samples) as the
-    # usual (n_samples, n_features). Furthermore, using a Fortran memory layout
+    # The dpex kernel expects X to be shaped as (n_features, n_samples) instead of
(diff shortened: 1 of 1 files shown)


ogrisel:
Also, I would rather name this variable `Xt` and the `centroids` variables `centroids_t` to make it more explicit that this kernel does not use the usual scikit-learn shape. However I wanted to keep this PR focused.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -156,7 +156,11 @@ def kmeans(
         number_of_load_iter=4,
     )
 
-    X = dpctl.tensor.from_numpy(np.ascontiguousarray(X).T, device=device)
+    # The dpex kernel expects X to be shaped as (n_features, n_samples) as the
+    # usual (n_samples, n_features). Furthermore, using a Fortran memory layout
```
ogrisel, line 160:
```suggestion
    # The dpex kernel expects X to be shaped as (n_features, n_samples) instead of
    # the usual (n_samples, n_features). Furthermore, using a Fortran memory layout
```

ogrisel:
Let's close to redo it on the new version of the k-means code.

PR #12, opened 2022-08-26
CI Improve pre-commit setup and run it on the CI

jjerphan:
This:
 - tune black correctly
 - use `isort` to sort import canonically
 - add a GitHub workflow to run format checks

Potentially we could have a single workflow with two jobs (one for those checks and another for the CI). What do you think?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     # TODO: remove pyopencl when dpctl exposes preferred_work_group_size_multiple
     # see https://github.com/IntelPython/dpctl/issues/886
     install_requires=["scikit-learn", "numba-dpex", "dpctl", "pyopencl"],
-    extras_require=[dict(benchmark=["scikit-learn-intelex, pandas"])],
+    extras_require=dict(benchmark=["scikit-learn-intelex", "pandas"]),
     packages=["sklearn_numba_dpex"],
     entry_points={
         "sklearn_engines": ["kmeans=sklearn_numba_dpex.cluster.kmeans:KMeansEngine"]
--- a/setup.py
+++ b/setup.py
@@ -1,45 +1,6 @@
 from setuptools import setup
 
 
-LONG_DESCRIPTION = """\
-TODO
-"""
-
-
 setup(
-    name="sklearn-numba-dpex",
-    maintainer="Olivier Grisel",
-    maintainer_email="olivier.grisel@ensta.org",
-    description="Computational Engine for scikit-learn based on numba-dpex",
-    license="BSD 3-Clause License",
-    url="https://github.com/ogrisel/sklearn-numba-dpex",
-    version="0.1.0",
-    long_description=LONG_DESCRIPTION,
-    classifiers=[
-        "Intended Audience :: Science/Research",
-        "Intended Audience :: Developers",
-        "License :: OSI Approved",
-        "Programming Language :: Python",
-        "Topic :: Software Development",
-        "Topic :: Scientific/Engineering",
-        "Development Status :: 4 - Beta",
-        "Operating System :: Microsoft :: Windows",
-        "Operating System :: POSIX",
-        "Operating System :: Unix",
-        "Operating System :: MacOS",
-        "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
-        "Programming Language :: Python :: Implementation :: CPython",
-    ],
-    python_requires=">=3.8",
-    # TODO: remove pyopencl when dpctl exposes preferred_work_group_size_multiple
-    # see https://github.com/IntelPython/dpctl/issues/886
-    install_requires=["scikit-learn", "numba-dpex", "dpctl", "pyopencl"],
-    extras_require=dict(benchmark=["scikit-learn-intelex", "pandas"]),
     packages=["sklearn_numba_dpex"],
-    entry_points={
-        "sklearn_engines": ["kmeans=sklearn_numba_dpex.cluster.kmeans:KMeansEngine"]
-    },
 )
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -37,11 +37,9 @@
     # TODO: remove pyopencl when dpctl exposes preferred_work_group_size_multiple
     # see https://github.com/IntelPython/dpctl/issues/886
     install_requires=["scikit-learn", "numba-dpex", "dpctl", "pyopencl"],
-    extra_requires=[dict(benchmark=["scikit-learn-intelex, pandas"])],
+    extra_require=[dict(benchmark=["scikit-learn-intelex, pandas"])],
```
ogrisel, line 40: Thanks for fixing this! I could not understand what was wrong myself!

sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -71,16 +72,16 @@ def make_fused_fixed_window_kernel(
...
     def fused_kmeans(
+        # fmt: off
         X_t,                               # IN READ-ONLY   (n_features, n_samples)
         current_centroids_t,               # IN             (n_features, n_clusters)
         centroids_half_l2_norm,            # IN             (n_clusters,)
         per_sample_pseudo_inertia,         # OUT            (n_samples,)
-        new_centroids_t_private_copies,    # OUT            (n_private_copies, n_features, n_clusters)
-        cluster_sizes_private_copies,      # OUT            (n_private_copies, n_clusters)
+        new_centroids_t_private_copies,    # OUT            (n_private_copies, n_features, n_clusters)  # noqa
+        cluster_sizes_private_copies,  # OUT            (n_private_copies, n_clusters)  # noqa
```
ogrisel, line 82: We could keep this comment consistently aligned with the others here.

```suggestion
        new_centroids_t_private_copies,    # OUT            (n_private_copies, n_features, n_clusters)  # noqa
        cluster_sizes_private_copies,      # OUT            (n_private_copies, n_clusters)  # noqa
```

setup.py
jjerphan: An extra character and the world is broken. :upside_down_face:

I find the inconsistency in the naming of `install_requires` and `extra_require` weird and error-prone as shown here.

sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -71,16 +72,16 @@ def make_fused_fixed_window_kernel(
...
     ):
-    # fmt: on
+        # fmt: on
```
ogrisel, line 84: Question: what does this stand for?

ogrisel (approved):
Thanks for setting this up. +1 for me if CI is green.

ogrisel:
> Potentially we could have a single workflow with two jobs (one for those checks and another for the CI). What do you think?

No strong opinion. I like having 2 different lines in the status summary of the PR on github.

sklearn_numba_dpex/kmeans/kernels.py
jjerphan: Rah, there is some hardness in having black ignore this block in the signature of functions.

jjerphan:
> No strong opinion. I like having 2 different lines in the status summary of the PR on github.

Good to know, and so do I, hence the current setup.

ogrisel:
```
/opt/venv/lib/python3.9/site-packages/setuptools/_distutils/dist.py:262: UserWarning: Unknown distribution option: 'extra_require'
```

Hum...

```
error: invalid command 'bdist_wheel'
```

is the `wheel` package needed as an explicit dependency?

jjerphan:
Hum, there is something inconsistent with the setup of the CI on `main`.

sklearn_numba_dpex/kmeans/kernels.py
fcharras: `fmt: off` and `fmt: on` what aimed at having black ignore the block. sad that it doesn't work :cry:

sklearn_numba_dpex/kmeans/kernels.py
fcharras: it was supposed to disable black within the block delimited by those tags

fcharras (approved):

setup.py
fcharras, line 40:
```suggestion
    extras_require=[dict(benchmark=["scikit-learn-intelex, pandas"])],
```
apparently there's an "s" on extra

fcharras:
Cool initiative very needed, thank you, LGTM when it's green, would be a plus to have black correctly ignores the signature blocks in the kernels .

The current issue seems to come from the file pyproject.toml .

fcharras:
I've looked a bit more into the `pyproject.toml` format and I think that your use of `[build-system]` is misguided, its use is closer to `setup_requires` than `install_requires`. If we want to configure the dependency tree of the package in this file the relevant config blocks are [dependencies/optional-dependencies](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dependencies-optional-dependencies).

Also we should not have several sources for the dependencies, it's either `setup.py`, `setup.cfg` or `pyproject.toml`. From the look of it, nowadays the motion is going towards `pyproject.toml` so let's use that exclusively if we can.

fcharras:
Regarding the current `workflow` file, why pass `--no-build-isolation` to `pip` rather than using the default ? one issue was that if not using build isolation pip expects some dependencies (like `wheels`, apparently bdist_wheels are a must have nowadays) to be pre-installed. So we can either explicitly install wheel manually or accept build isolation, since we install only a single package it doesn't make much of a difference. (also, maybe adding wheels as a `setup_requires` would work)

fcharras:
I'll add an additional commit that moves all the setup information to the `pyproject,toml`

jjerphan:
> Regarding the current workflow file, why pass --no-build-isolation to pip rather than using the default ?

Habits from local development. This is indeed not needed here.

fcharras:
So I've pushed more things into the `pyproject.toml`. 

Honestly I think it's OK to have configuration spread into `setup.py`, `pyproject.toml` and `setup.cfg`, Some of those configs seems to still have some specs that are not shared with others.

What is not OK would have to be a same config repeated in different places (and worse, with different values).

Python packaging being what it is.

ogrisel:
Since this is a pure python project (no compiled dependencies) we could probably put everything in `pyproject.toml` with the `flit` packager I think. But this looks already like a net improvement. Let's merge since CI is green.

PR #14, opened 2022-09-01
Reduce the number of duplicated codes in kmeans kernel

fcharras:
We have currently [this kernel](https://github.com/soda-inria/sklearn-numba-dpex/blob/main/sklearn_numba_dpex/kmeans/kernels.py#L75) and [this kernel](https://github.com/soda-inria/sklearn-numba-dpex/blob/main/sklearn_numba_dpex/kmeans/kernels.py#L344) that have duplicated code, meaning if changes happen on one kernel probably it should be propagated to the other one, which is not very good for maintaining.

But `numba_dpex` kernels can only be refactored using `numba_dpex.func`-decored [device functions](https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html) and those have a limited set of features, e.g. you can't use barriers inside or create arrays, those particular steps must be in the body of the kernel. In this case those parts are going to stay replicated anyway and we try to factor the parts in-between only.

I also attempted this because for implementing `sklearn` predict we might want a kernel that only returns labels (inertia output is not needed), and for steps taken when managing empty clusters we might only want distances, so that would be two more occurences of duplicated lines.

Given how the kernels doesn't factor well maybe it's better to keep the duplicated version, and stick with only one kernel that returns all outputs, that we discard afterward if not needed ?

### DIFF
--- a/sklearn_numba_dpex/kmeans/kernels.py
+++ b/sklearn_numba_dpex/kmeans/kernels.py
@@ -58,7 +58,12 @@ def make_kmeans_fixed_window_kernels(
     )
 
     _accumulate_dot_products = _make_accumulate_dot_products_kernel_func(
-        n_samples, n_features, centroids_window_height, window_n_centroids, False, dtype
+        n_samples,
+        n_features,
+        centroids_window_height,
+        window_n_centroids,
+        with_X_l2_norm=False,
+        dtype=dtype,
     )
 
     n_windows_per_feature = math.ceil(n_clusters / window_n_centroids)
@@ -91,16 +96,16 @@ def make_kmeans_fixed_window_kernels(
     # the following kernel definition contains a lot of inline comments but those
     # comments are not repeated in the similar patterns in the other kernels
     @dpex.kernel
+    # fmt: off
     def fused_kmeans(
-        # fmt: off
         X_t,                               # IN READ-ONLY   (n_features, n_samples)
         current_centroids_t,               # IN             (n_features, n_clusters)
         centroids_half_l2_norm,            # IN             (n_clusters,)
         per_sample_pseudo_inertia,         # OUT            (n_samples,)
         new_centroids_t_private_copies,    # OUT            (n_private_copies, n_features, n_clusters)  # noqa
-        cluster_sizes_private_copies,  # OUT            (n_private_copies, n_clusters)  # noqa
+        cluster_sizes_private_copies,      # OUT            (n_private_copies, n_clusters)  # noqa
     ):
-        # fmt: on
+    # fmt: on
         """One full iteration of LLoyd's k-means.
 
         The kernel is meant to be invoked on a 1D grid spanning the data samples of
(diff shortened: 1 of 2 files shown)


sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -40,22 +40,31 @@ def make_fused_fixed_window_kernel(
...
+    )
+
+    _update_closest_centroid = _make_update_closest_centroid_kernel_func(
+        window_n_centroids
     )
 
-    window_n_features = centroids_window_height
+    _accumulate_dot_products = _make_accumulate_dot_products_kernel_func(
+        n_samples, n_features, centroids_window_height, window_n_centroids, False, dtype
+    )
```
jjerphan, line 62:
```suggestion
    _accumulate_dot_products = _make_accumulate_dot_products_kernel_func(
        n_samples,
        n_features,
        centroids_window_height,
        window_n_centroids,
        with_X_l2_norm=False,
        dtype=dtype,
    )
```

sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -466,7 +396,189 @@ def assignment(
...
+    zero = dtype(0.0)
+    inf = dtype(math.inf)
+
+    @dpex.func
+    def _initialize_window_of_centroids(
+        local_work_id,
+        first_centroid_idx,
+        centroids_half_l2_norm,
+        centroids_window_half_l2_norm,
+        dot_products,
```
jjerphan, line 424:
```suggestion
        local_work_id,
        first_centroid_idx,
        centroids_half_l2_norm,         # IN
        centroids_window_half_l2_norm,  # OUT
        dot_products,                   # OUT
```

sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -466,7 +396,189 @@ def assignment(
...
+    )
+
+    @dpex.func
+    def _load_window_of_centroids_and_features(
+        first_feature_idx,
+        loading_centroid_idx,
+        window_loading_centroid_idx,
+        window_loading_feature_offset,
+        current_centroids_t,
+        centroids_window,
```
jjerphan, line 455: I think knowing which arguments is a tensor and if it gets modified really help both reading and understanding implementations. What do you think?
```suggestion
        current_centroids_t,    # IN
        centroids_window,       # OUT
```

The same suggestion apply to other kernels' signature.

sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -466,7 +396,189 @@ def assignment(
...
+    return _initialize_window_of_centroids, _load_window_of_centroids_and_features
+
+
+def _make_accumulate_dot_products_kernel_func(
+    n_samples, n_features, window_n_features, window_n_centroids, with_X_l2_norm, dtype
+):
+    # !!!: this is really ugly but I don't think numba.func can support more than this
+    # level of factorization. I tried using a const bool to swtich X_l2_norm computation
+    # on and off coupled with explicitly passing the `signature` argunment, but to no
+    # result (the JIT will fail).
```
jjerphan, line 493: If it can be better implemented in the future using new features, then let's do it.
As of now, I think it is manageable to maintain the switch as is.

jjerphan (approved):
Thank you, @fcharras.

I am +0.5 for this factorisation given the current features set of numba_dpex.

If there is a way which arguments are tensor and which gets modified in place in kernels (be it a feature or a convention), then I am +10.

fcharras:
LGTM then, i applied the suggested changes plus some extras for readability, TY for review @jjerphan

jjerphan:
You are welcome, @fcharras.

PR #15, opened 2022-09-20
Implement inference methods (predict, transform, and score)

fcharras:


fcharras:
So this is the private API that was missing for inference methods (KMeans.transform. KMeans.predict, and KMeans.score) + unit tests.

Also I've dispatched the main kernels in several distinct files, it does not fix the issues with repeated code lines  but I think it feels lighter this way.

sklearn_numba_dpex/kmeans/kernels.py
```diff
@@ -1,942 +0,0 @@
-import math
-from functools import lru_cache
-
-import dpctl
-import numba_dpex as dpex
-
-# General note on kernel implementation
```
jjerphan, line 7: Is it possible to do another PR which would precedes this PR for the removal of this file and the migration of its content to the files in `sklearn_numba_dpex/kmeans/kernels`? This would reduce the diff of this PR and I think ease reviews (I think having at max ± 500 diffs is a good rule of thumb for faster integration).

jjerphan (commented):
Thank you, @fcharras.

Here is a high level pass with general suggestions to make this PR easier to review.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -166,7 +170,7 @@ def __init__(
                     f"support it."
                 )
 
-    def __call__(
```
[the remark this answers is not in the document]
fcharras: Let's rename it KMeansDriver then ?

benchmark/kmeans.py
```diff
@@ -217,14 +217,14 @@ def benchmark_data_initialization(random_state=random_state, n_clusters=n_cluste
...
             slow=True,
         )
 
         kmeans_timer.timeit(
             LLoydKMeansDriver(
                 device="gpu", dtype=dtype, work_group_size_multiplier=multiplier
-            ),
+            ).lloyd,
             name=f"Kmeans numba_dpex lloyd GPU (work_group_size_multiplier={multiplier})",
         )
```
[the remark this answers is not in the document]
fcharras: That is out of scope of this PR i'd prefer not address that here, it is expected that benchmarks will be iterated on later (adding new dataset, a better grid search, etc)

sklearn_numba_dpex/kmeans/kernels/_base_kmeans_kernel_funcs.py
```diff
@@ -0,0 +1,195 @@
...
+    def _initialize_window_of_centroids(
+        local_work_id,                  # PARAM
+        first_centroid_idx,             # PARAM
+        centroids_half_l2_norm,         # IN
+        centroids_window_half_l2_norm,  # OUT
+        dot_products,                   # OUT
+    ):
+    # fmt: on
+        # initialize the partial pseudo inertia for each of the window_n_centroids
+        # centroids in the window
```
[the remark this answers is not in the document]
fcharras: i'll apply the 3 following nits in #18 where it now belongs

sklearn_numba_dpex/kmeans/kernels.py
fcharras: While I agree with your rule of thumbs for diffs, i don't think it should apply here since the diff was artificially high because of copy/paste, even if they appear in the diff I don't think the internals of the copy/pasted blocks should be reviewed, they have not been edited, only moved left and right.

Even after rebasing on #18 it remains fairly high because of duplicated code that can't be refactored because known limitations with numba_dpex.

fcharras:
Thank you for looking at this PR @jjerphan I've addressed your requests here and in #18

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+        X_t, centroids_t = self._load_transposed_data_to_device(X, centers)
+
+        centroids_half_l2_norm = dpctl.tensor.empty(
+            n_clusters, dtype=dtype, device=self.device
+        )
+        assignments_idx = dpctl.tensor.empty(
+            n_samples, dtype=np.uint32, device=self.device
+        )
+
+        half_l2_norm(centroids_t, centroids_half_l2_norm)
```
jjerphan, line 472: This adds a `_kernel` suffix to `half_l2_norm` to match the kernels' naming convention.
```suggestion
        half_l2_norm_kernel = make_half_l2_norm_2d_axis0_kernel(
            size0=n_features,
            size1=n_clusters,
            work_group_size=work_group_size,
            dtype=dtype,
        )

        X_t, centroids_t = self._load_transposed_data_to_device(X, centers)

        centroids_half_l2_norm = dpctl.tensor.empty(
            n_clusters, dtype=dtype, device=self.device
        )
        assignments_idx = dpctl.tensor.empty(
            n_samples, dtype=np.uint32, device=self.device
        )

        half_l2_norm_kernel(centroids_t, centroids_half_l2_norm)
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+            work_group_size=work_group_size,
+            dtype=dtype,
+        )
+
+        reduce_inertia = make_sum_reduction_1d_kernel(
+            size=n_samples,
+            work_group_size=work_group_size,
+            device=self.device,
+            dtype=dtype,
+        )
```
jjerphan, line 529: Same suggestion regarding adding a suffix for kernels' names.
```suggestion
        half_l2_norm_kernel = make_half_l2_norm_2d_axis0_kernel(
            size0=n_features,
            size1=n_clusters,
            work_group_size=work_group_size,
            dtype=dtype,
        )

        reduce_inertia_kernel = make_sum_reduction_1d_kernel(
            size=n_samples,
            work_group_size=work_group_size,
            device=self.device,
            dtype=dtype,
        )
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+        X_t, centroids_t = self._load_transposed_data_to_device(X, centers)
+
+        centroids_half_l2_norm = dpctl.tensor.empty(
+            n_clusters, dtype=dtype, device=self.device
+        )
+        per_sample_inertia = dpctl.tensor.empty(
+            n_samples, dtype=dtype, device=self.device
+        )
+
+        half_l2_norm(centroids_t, centroids_half_l2_norm)
```
jjerphan, line 540:
```suggestion
        half_l2_norm_kernel(centroids_t, centroids_half_l2_norm)
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+
+        compute_inertia_fixed_window_kernel(
+            X_t,
+            centroids_t,
+            centroids_half_l2_norm,
+            per_sample_inertia,
+        )
+
+        inertia = dpctl.tensor.asnumpy(reduce_inertia(per_sample_inertia))[0]
```
jjerphan, line 549: Why do we need to take the first returned element solely here?
What is the instruction returning?
```suggestion
        inertia = dpctl.tensor.asnumpy(reduce_inertia_kernel(per_sample_inertia))[0]
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+
+    def get_euclidean_distances(
+        self, X, Y=None, *, Y_norm_squared=None, squared=False, X_norm_squared=None
+    ):
+        """This call is expected to accept the same inputs than sklearn's private
+        euclidean_distances and returns euclidean distances of each sample to each
+        cluster center
+        """
+        if squared:
+            raise NotImplementedError("Only squared=False is allowed")
```
jjerphan, line 561: Why is it the case exactly?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+            n_features,
+            n_clusters,
+            preferred_work_group_size_multiple=self.preferred_work_group_size_multiple,
+            centroids_window_width_multiplier=self.centroids_window_width_multiplier,
+            centroids_window_height=self.centroids_window_height,
+            work_group_size=work_group_size,
+            dtype=dtype,
+        )
+
+        half_l2_norm = make_half_l2_norm_2d_axis0_kernel(
```
jjerphan, line 585:
```suggestion
        half_l2_norm_kernel = make_half_l2_norm_2d_axis0_kernel(
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+        )
+
+        X_t, Y_t = self._load_transposed_data_to_device(X, Y)
+
+        Y_half_l2_norm = dpctl.tensor.empty(n_clusters, dtype=dtype, device=self.device)
+        euclidean_distances_t = dpctl.tensor.empty(
+            (n_clusters, n_samples), dtype=dtype, device=self.device
+        )
+
+        half_l2_norm(Y_t, Y_half_l2_norm)
```
jjerphan, line 599:
```suggestion
        half_l2_norm_kernel(Y_t, Y_half_l2_norm)
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+        self,
+        X,
+        sample_weight,
+        x_squared_norms,
+        centers,
+        n_threads=1,
+        return_inertia=True,
+    ):
+        """This call is expected to accept the same inputs than sklearn's private
+        _labels_inertia and returns (labels, None)
```
jjerphan, line 432:
```suggestion
        """This call is expected to accept the same inputs than
        `sklearn.cluster._kmeans._labels_inertia` while solely computing
        the samples' labels, hence returning (labels, None).
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +419,194 @@ def __call__(
...
+        self,
+        X,
+        sample_weight,
+        x_squared_norms,
+        centers,
+        n_threads=1,
+        return_inertia=True,
+    ):
+        """This call is expected to accept the same inputs than sklearn's private
+        _labels_inertia and returns (None, inertia)
```
jjerphan, line 493:
```suggestion
        """This call is expected to accept the same inputs than
        `sklearn.cluster._kmeans._labels_inertia` while solely computing
        the samples' inertia, hence returning (None, inertia).
```

jjerphan (commented):
A pass on the tests and the implementations of the methods.

I still need to review `sklearn_numba_dpex/kmeans/kernels/compute_{euclidean_distances,inertia,labels}.py`.

sklearn_numba_dpex/kmeans/kernels.py
jjerphan: > While I agree with your rule of thumbs for diffs, i don't think it should apply here since the diff was artificially high because of copy/paste, even if they appear in the diff I don't think the internals of the copy/pasted blocks should be reviewed, they have not been edited, only moved left and right.

I agree, but to me, having blocks of code moved around without changes still hardened reviews. PR like #18 highly help in this regards.

sklearn_numba_dpex/kmeans/drivers.py
fcharras: We monkey patch sklearn internals with this function in places where it's always called with `squared=False`.

This check ensures that this assumption holds.

Could as well be `RuntimeError` or `ValueError`.

I didn't try to make the message more explicit on purpose: all this monkey-patching salad is going to be removed in a soon to come PR in favor of the more conventional plugin interface.

sklearn_numba_dpex/kmeans/drivers.py
fcharras: no actually I'm wrong: the current code is good, `reduce_inertia_kernel(per_sample_inertia)` is a 1-sized array

sklearn_numba_dpex/kmeans/drivers.py
jjerphan: OK, then if supporting `squared=True` is possible but useless, I think it is worth adding an inline comment to mention it. For instance:
```suggestion
        if squared:
            # This implementation is used to monkey patch sklearn private
            # functions' which are always called with `squared=False`.
            # It's possible to support `squared=True`, but useless in this
            # context, hence we raise an error.
            raise NotImplementedError("Only squared=False is allowed")
```

sklearn_numba_dpex/kmeans/drivers.py
jjerphan: In this case, I think having a comment indicates that it is a 1-sized array whose scalar is accessed via `[0]` would be useful for readers.

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -0,0 +1,147 @@
...
+                dot_products,
+            )
+
+            dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+
+            first_centroid_idx += window_n_centroids
+
+        # No update step, only store min_idx in the output array
+        if sample_idx >= n_samples:
+            return
```
jjerphan, line 142: Shouldn't this be moved at before for-loops?

sklearn_numba_dpex/kmeans/kernels/compute_inertia.py
```diff
@@ -0,0 +1,165 @@
...
+                centroids_window_half_l2_norm,
+                dot_products,
+            )
+
+            dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+
+            first_centroid_idx += window_n_centroids
+
+        if sample_idx >= n_samples:
+            return
```
jjerphan, line 160: Shouldn't this be moved at before for-loops?

jjerphan (approved):
LGTM. Just a remark.

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
fcharras: ah this is a good question and actually a subtle point that would benefit from a comment, IIRC I found that an undesirable work item (such as `sample_idx >= n_samples`) still has to reach the barriers, else the kernel will deadlock.

In general kernels with such code

```python
If some_condition:
    do_sth()
    dpex.barrier(...)
else:
    maybe_do_sth_else()
```

will deadlock, and must be written like this:

```python
If some_condition:
    do_sth()
else:
    maybe_do_sth_else().
    # no barrier

dpex.barrier(...)
```

jjerphan: OK, I think this better be made explicit in a comment as you suggest.

fcharras:
Rebased to show co authorship in review commits I think it worked ?

jjerphan:
Yes, it worked. Commits are marked as unverified when their (co-)author usually GPG-sign them.

PR #18, opened 2022-09-22
refacto: define important kernels in separate files

fcharras:
The `kernels.py` file is broken down into several files in the `kernels` folder, in preparation of the PR https://github.com/soda-inria/sklearn-numba-dpex/pull/15 that is going to add more kernels in additional files.

There's also a little bit of anticipated refactoring in the file driver.py.

This PR is nothing more than a set of copy/paste diff (please review how code blocks are moved but not the line by line content, which has not been edited)


sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -356,7 +346,7 @@ def __call__(
             reset_centroids_private_copies(new_centroids_t_private_copies)
 
             # TODO: implement special case where only one copy is needed
-            fixed_window_fused_kernel(
+            fused_lloyd_fixed_window_single_step(
```
jjerphan, line 349: If the other suggestion is approved, this one also must.
```suggestion
            fused_lloyd_fixed_window_single_step_kernel(
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -180,23 +181,21 @@ def __call__(
...
+        ) = self._check_inputs(X, centers_init)
 
         # Create a set of kernels
         (
             n_centroids_private_copies,
-            fixed_window_fused_kernel,
-            fixed_window_assignment_kernel,
-        ) = make_kmeans_fixed_window_kernels(
+            fused_lloyd_fixed_window_single_step,
+        ) = make_lloyd_single_step_fixed_window_kernel(
```
jjerphan, line 198: Has the _kernel suffix be removed intentionally here? Also, there is a discrepancy w.r.t the name of the factory. Conventionally, I would have expected those two names:

```suggestion
            fused_lloyd_fixed_window_single_step_kernel,
        ) = make_fused_lloyd_fixed_window_single_step_kernel(
```

Does this makes sense?

fcharras: I thought that the readability of the loop in this function (the "high level code") is better with names that only describe what the function does functionally (thinking of inputs/outputs) rather than other characteristics (like its nature of being a kernel)

jjerphan: This makes sense to me. Do you think it's possible to have a name which covers both aspect (i.e. what this performs and it being a kernel)?

fcharras: no :shrug: but if you think it's better i'll add _kernel prefix everywhere

jjerphan: Either is fine with me.

PR #19, opened 2022-09-22
Detect empty clusters during lloyd algorithm and issue a warning

fcharras:
This is a first step toward appropriate dealing of empty clusters.

In this PR we detect empty clusters when it happens during the lloyd algorithm and issue a warning when it happens.

Future work will implement the exact behavior that we have in scikit learn. In the short term I want to prioritize the plugin interface and probably will come to empty clusters only afterward (or maybe somebody else want to give it a try before me)

But since I had this part pending I think we could merge it already.

### DIFF
--- a/sklearn_numba_dpex/kmeans/drivers.py
+++ b/sklearn_numba_dpex/kmeans/drivers.py
@@ -316,6 +316,7 @@ class LLoydKMeansDriver:
             n_clusters, dtype=np.uint32, device=self.device
         )
 
+        # nb_empty_clusters_ is a scalar handled in kernels via a one-element array.
         nb_empty_clusters = dpctl.tensor.empty(1, dtype=np.int32, device=self.device)
 
         # The loop
--- a/sklearn_numba_dpex/kmeans/drivers.py
+++ b/sklearn_numba_dpex/kmeans/drivers.py
@@ -11,6 +11,7 @@ from sklearn_numba_dpex.kmeans.kernels import (
     make_lloyd_single_step_fixed_window_kernel,
     make_compute_labels_inertia_fixed_window_kernel,
     make_centroid_shifts_kernel,
+    make_reduce_centroid_data_kernel,
     make_initialize_to_zeros_1d_kernel,
     make_initialize_to_zeros_2d_kernel,
     make_initialize_to_zeros_3d_kernel,
@@ -18,8 +19,6 @@ from sklearn_numba_dpex.kmeans.kernels import (
     make_broadcast_division_1d_2d_kernel,
     make_half_l2_norm_2d_axis0_kernel,
     make_sum_reduction_1d_kernel,
-    make_sum_reduction_2d_axis0_kernel,
-    make_sum_reduction_3d_axis0_kernel,
 )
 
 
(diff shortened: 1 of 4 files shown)


sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -58,6 +59,66 @@ def centroid_shifts(
     return centroid_shifts[global_size, work_group_size]
 
 
+@lru_cache
+def make_reduce_centroid_data(
```
fcharras, line 63: this kernel is the fusion of `make_sum_reduction_2d_axis0_kernel` and `make_sum_reduction_3d_axis0_kernel` + some extra instructions to register empty clusters.

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -58,6 +59,68 @@ def centroid_shifts(
     return centroid_shifts[global_size, work_group_size]
 
 
+@lru_cache
+def make_reduce_centroid_data_kernel(
```
fcharras, line 63: this kernel is the fusion of `make_sum_reduction_2d_axis0_kernel` and `make_sum_reduction_3d_axis0_kernel` + some extra instructions to register empty clusters.

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -75,6 +77,60 @@ def test_kmeans_fit_same_results(dtype):
...
+    _fail_if_no_dtype_support(pytest.xfail, dtype)
+
+    random_seed = 42
+    X, _ = make_blobs(random_state=random_seed)
+    X = X.astype(dtype)
+
+    n_clusters = inspect.signature(KMeans).parameters["n_clusters"].default
+    rng = default_rng(random_seed)
+    init = np.array(rng.choice(X, n_clusters, replace=False), dtype=np.float32)
+    init[0] = np.finfo(np.float32).max
```
jjerphan, line 91:
```suggestion
    # Create an outlier centroid in initial centroids
    # to have an empty cluster.
    init[0] = np.finfo(np.float32).max
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -320,6 +312,11 @@ def __call__(
             dtype=dtype,
             device=self.device,
         )
+        empty_clusters_list = dpctl.tensor.empty(
+            n_clusters, dtype=np.uint32, device=self.device
+        )
+
+        nb_empty_clusters = dpctl.tensor.empty(1, dtype=np.int32, device=self.device)
```
jjerphan, line 319: Nit: attempt to make the `[0]`-accesses explicit.
```suggestion
        # nb_empty_clusters_ is a scalar handled in kernels via a one-element array.
        nb_empty_clusters = dpctl.tensor.empty(1, dtype=np.int32, device=self.device)
```

jjerphan (approved):
LGTM.

Here are minor nitpicks.

fcharras:
thanks for the review! will apply the suggestion and merge

PR #20, opened 2022-09-26
High-level API for registering the engine with scikit-learn's entry point

fcharras:
This branch is intended to be used along with the sklearn branch in https://github.com/scikit-learn/scikit-learn/pull/24497


pyproject.toml
```diff
@@ -54,7 +54,7 @@ benchmark = [
 
 
 [project.entry-points.sklearn_engines]
-kmeans = "sklearn_numba_dpex.cluster.kmeans:KMeansEngine"
+kmeans = "sklearn_numba_dpex.kmeans.drivers:KMeansSklearnEngine"
```
ogrisel, line 57: Maybe this would feel more natural to me (avoid putting sklearn everywhere).

```suggestion
kmeans = "sklearn_numba_dpex.kmeans.engine:KMeansEngine"
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -31,6 +33,21 @@ def _check_power_of_2(e):
     return e
 
 
+class KMeansSklearnEngine(KMeansCythonEngine):
+    def kmeans_single(self, X, sample_weight, centers_init):
+        print("Using sklearn_numba_dpex engine for fitting KMeans...")
```
ogrisel, line 38: I would rather not have anything printed out by default.

ogrisel (approved):
LGTM as a first step before implementing the other engine methods.

ogrisel:
We also need to run the tests on the CI against the `wip-engines` branch and/or skip tests if the default Cython `KMeans` engine cannot be found in scikit-learn.

fcharras:
OK so this appears to work without too much boilerplate: if all the features that are tested in scikit-learn but are not supported by the plugin ends up raising `sklearn_numba_dpex.exceptions.NotImplementedError`, using the `pytest` command like it's described in the `README` will mark all the `sklearn` tests that cover unsupported features as "xfailed".

Remaining points i'd like to address in this PR:

- [ ] add the relevant `sklearn` test suites in our automated testing pipeline
- [ ] all the failures are currently because of unsupported features (float64 dtype without hardware support / elkan / sparse inputs), let's improve input validation until all those tests correctly raise `NotImplementedError`

And somewhat urgent work that could be done in a next PR would be to support non uniform `sample_weight` because almost all tests in sklearn have non uniform sample weights.

The `-p` and the `--pyargs` option could be added to `setup.cfg` so that a single command without arguments can start all the tests it it a good pytest practice ?

pyproject.toml
```diff
@@ -54,7 +54,11 @@ benchmark = [
 
 
 [project.entry-points.sklearn_engines]
-kmeans = "sklearn_numba_dpex.cluster.kmeans:KMeansEngine"
+kmeans = "sklearn_numba_dpex.kmeans.engine:KMeansSklearnEngine"
```
ogrisel, line 57: Let's also remove `Sklearn` from the engine class name.

```suggestion
kmeans = "sklearn_numba_dpex.kmeans.engine:KMeansEngine"
```

is already long enoug.

README.md
```diff
@@ -179,6 +179,23 @@ See the `sklearn_numba_dpex/tests` folder for example usage.
...
+To run the tests run the following:
+
+```
+pytest sklearn_numba_dpex/
+```
+
+To run the `scikit-learn` tests with the `sklearn_numba_dpex` engine you can run the following:
+
+```
+cd ./sklearn_tests
```
ogrisel, line 193: We don't need this folder anymore:

```suggestion
```

ogrisel (approved):
Nice plugin trick.

Could you please install the `wip-engines` branch from scikit-learn prior to this step in:

https://github.com/soda-inria/sklearn-numba-dpex/blob/b859d2ed895e9e4286f0acb205d3c34e48782721/.github/workflows/run_tests.yaml#L18-L19

otherwise LGTM besides the final comments:

ogrisel:
> The -p and the --pyargs option could be added to setup.cfg so that a single command without arguments can start all the tests it it a good pytest practice ?

I think it's better to keep it explicit. Otherwise it's a bit too magic to my taste.

ogrisel:
Maybe the plugin could be renamed to "xfail-not-implemented" or something more descriptive (not sure if `-` signs are fine in pytest plugin names).

README.md
```diff
@@ -179,6 +179,23 @@ See the `sklearn_numba_dpex/tests` folder for example usage.
...
+pytest -p sklearn_numba_dpex --pyargs sklearn.cluster.tests.test_k_means
+```
+
+(change the `--pyargs` option accordingly to select other test suites). Note that the `sklearn_numba_dpex` pytest plugin will automatically activate the `sklearn_numba_dpex` engine for all tests, and tests that cover unsupported features (that trigger `sklearn_numba_dpex.exceptions.NotImplementedError`) will be automatically marked as xfailed.
```
jjerphan, line 197: Julien being nitpicky.
```suggestion
(change the `--pyargs` option accordingly to select other test suites).
Note that the `sklearn_numba_dpex` pytest plugin will automatically activate the `sklearn_numba_dpex` engine for all tests.
Also note that tests covering unsupported features (that trigger `sklearn_numba_dpex.exceptions.NotImplementedError`) will be automatically marked as xfailed.
```

pyproject.toml
```diff
@@ -54,7 +54,11 @@ benchmark = [
...
+
+[project.entry-points.pytest11]
```
jjerphan, line 59: What does `pytest11` stands for, here?

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -0,0 +1,19 @@
+try:
+    from sklearn.cluster._kmeans import KMeansCythonEngine
+except ImportError:
+    KMeansCythonEngine = object
```
jjerphan, line 4: Why do we need `KMeansSklearnEngine` to subclass `KMeansCythonEngine`?

jjerphan (commented):
Thank you, @fcharras. Here are a few comments.

fcharras:
@ogrisel 

> I think it's better to keep it explicit. Otherwise it's a bit too magic to my taste.

agree

> Maybe the pytest plugin could be renamed to "xfail-not-implemented" or something more descriptive (not sure if `-` signs are fine in pytest plugin names).

The pytest plugin is specific to `sklearn_numba_dpex` engines (since it activates its engine and catches `sklearn_numba_dpex.exceptions.NotImplementedError`) so I think the name must relate to `sklearn_numba_dpex` rather than being a generic one, and anything other than just `sklearn_numba_dpex` sounds too long, in fact I like having just `sklearn_numba_dpex`.

fcharras:
It's starting to look good, a few remaining tests won't pass on my machine for the following reasons:

- 2 or 3 tests where `allclose` fail because of slight numeric errors (just above the thresholds), probably because on my machine it uses the gpu device but when dealing with float64 it fallbacks to float32 because float64 is not supported by my gpu device
- verbosity tests, to pass those we need to mimics the prints in the default engine
- a convergence test, because sklearn have an additional step to detect early convergence
- the test on empty clusters

About installing the wip-engines branch in the CI, it requires building sklearn from source so it's kind of slow.

Currently the CI fail with an issue I can't reproduce locally, pytest fails to load properly.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -665,23 +725,31 @@ def _set_dtype(self, X, centers_init):
...
-        return X, centers_init, dtype
+        return X, centers_init, compute_dtype, output_dtype
 
-    def _check_inputs(self, X, cluster_centers):
-        X, cluster_centers, dtype = self._set_dtype(X, cluster_centers)
+    def _check_inputs(self, X, cluster_centers, sample_weight):
+        if (sample_weight is not None) and ((sample_weight != sample_weight[0]).any()):
+            raise FeatureNotAvailableError(
+                "The sklearn_numba_dpex engine for KMeans does not support non-uniform"
+                " sample weights."
```
ogrisel, line 747: Maybe we could be explicit that this is a planned feature and include the URL of an issue tracking the implementation of this feature.

ogrisel (commented):
Thanks @fcharras. Here is a path of feedback.

sklearn_numba_dpex/testing/sklearn_pytest_plugin.py
```diff
@@ -0,0 +1,16 @@
...
+
+def pytest_runtest_call(item):
+    with config_context(engine_provider="sklearn_numba_dpex"):
+        try:
+            item.runtest()
+        except FeatureNotAvailableError:
+            xfail(
+                reason="This test cover features that are not supported by the "
+                "engine provided by sklearn_numba_dpex."
+            )
```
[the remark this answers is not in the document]
ogrisel: It's possible to pass commandline argument with `pytest_addoption`.

pyproject.toml
ogrisel: I had the same question. Is this because this is version 11 of the plugin API of pytest?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -163,7 +167,7 @@ def __init__(
                 raise ValueError(f"Valid types are float64, float32, but got f{dtype}")
             self.dtype = dtype
             if (self.dtype == np.float64) and not self.has_aspect_fp64:
-                raise RuntimeError(
+                raise FeatureNotAvailableError(
                     f"Computations with precision f{self.dtype} has been explicitly "
                     f"requested to the KMeans driver but the device {dpctl_device.name} does not "
                     f"support it."
```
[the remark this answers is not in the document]
ogrisel: I opened https://github.com/soda-inria/sklearn-numba-dpex/issues/21 to track this latter point.

fcharras:
So I pushed simultaneously on https://github.com/scikit-learn/scikit-learn/pull/24497 and here to:
- move the plugin system to sklearn (see the README about what changes when invoking pytest (not much)) amd nove the exception to sklearn.exceptions
- expose predict/score/transform methods on the default kmeans engine and on the sklearn_numba_dpex engine
- update tests accordingly

(NB: the benchmark script is broken now, needs a commit to fix it, i.e. adding a private api to access the parameters of the KMeansDriver when using it from within a sklearn plugin)

fcharras:
It think we could merge with this state ? 

In my opinion this PR lays the last stone of the POC, feature-wise the KMeans is almost complete and we have a nice plugin system.

jjerphan:
To build on top of @betatim's remark in https://github.com/scikit-learn/scikit-learn/pull/24497#issuecomment-1259181963, we probably need to clarify the semantic of "plugin" and "engines" first, don't you think?

Intuitively, I see "plugins" as be packages which could change or extend scikit-learn in various ways following an clear interface/process. "Engines" would be a set of (mostly computational) alternative implementations based on a particular set of technology (for a particular set of targeted hardware).

Engines would be part of some plugins, but plugins without engines could also exist and be supported.

Which understanding of plugins/engines do you have?

fcharras:
I don't have a strong opinion on this, it would help having examples of other software or codebases that use such wordings, from what I've read about the different opinions so far `engine` sounds sensible.

> we probably need to clarify [...]  _first_

I would'nt say that, it only takes a global find/replace to fix the naming only when the branch `wip-engines` is out of draft phase and want to claim stability/reliability, if this point is to request more discussion it can continue in parallel to the development for now.

I agree it's important to choose it right at the end though since that's mostly the wording the public and the users will read in the documentation and remember the most about this project.

fcharras:
The benchmark script is fixed in this new PR https://github.com/soda-inria/sklearn-numba-dpex/pull/23 that targets the branch of the current PR.

The two branch should be merged simultaneously to avoid breaking the benchmark script on main at any point.

.github/workflows/run_tests.yaml
```diff
@@ -15,14 +15,18 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
-    - name: Install sklearn-numba-dpex
-      run: pip install .
+      # TODO: replace development branch for engines in scikit learn with a fixed commit
+      # commit when it's stable / remove this step when it's officially released
+    - name: Install pytest, @ogrisel sklearn branch "wip-engines", and sklearn-numba-dpex
+      run: pip install pytest -e git+https://github.com/ogrisel/scikit-learn.git@wip-engines#egg=scikit-learn -e .
```
jjerphan, line 21: :+1:

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -131,12 +133,12 @@ def __init__(
             or device_params.preferred_work_group_size_multiple
         )
 
+        # So far the best default value has been empirically found to be 2 so we use
+        # this default.
+        # TODO: when it's available in dpctl, use the `max_group_size` attribute
+        # exposed by the kernel instead ?
```
jjerphan, line 139: IIRC, `numba_dpex` has a function to get the optimal values for `work_group_size` and similar variables; is this right, @oleksandr-pavlyk?

jjerphan (approved):
LGTM.

fcharras:
> Looks like pytest plugins does not work well for a package installed from source in non-editable mode

fixed in https://github.com/scikit-learn/scikit-learn/pull/24497/commits/70e4ecfdc9dcc57622339d095028e764c484c674

jjerphan:
> I would'nt say that, it only takes a global find/replace to fix the naming only when the branch wip-engines is out of draft phase and want to claim stability/reliability, if this point is to request more discussion it can continue in parallel to the development for now.

OK. LGTM then.

fcharras:
TY @jjerphan and @ogrisel for the reviews!!

PR #23, opened 2022-09-29
Fix the benchmark command (using the engine API)

fcharras:
This PR fixes the benchmark that breaks in the soon-to-merge PR https://github.com/soda-inria/sklearn-numba-dpex/pull/20

The two PRs should be merged simultaneously to avoid a regression on the benchmark script on main.

### DIFF
--- a/benchmark/ext_helpers/daal4py.py
+++ b/benchmark/ext_helpers/daal4py.py
@@ -66,7 +66,7 @@ def daal4py_kmeans_single(self, X, sample_weight, centers_init):
 # from daal4py.sklearn.cluster._k_means_0_23 import _daal4py_k_means_fit
 
 
-# def kmeans_daal4py_kmeans_singlesingle(self, X, sample_weight, centers_init):
+# def kmeans_daal4py_kmeans_single(self, X, sample_weight, centers_init):
 
 #     (
 #         best_cluster_centers,
--- a/benchmark/ext_helpers/daal4py.py
+++ b/benchmark/ext_helpers/daal4py.py
@@ -1,28 +1,56 @@
-from daal4py.sklearn.cluster._k_means_0_23 import KMeans
+import sklearn
+import warnings
 
 
-def kmeans(
-    X,
-    sample_weight,
-    centers_init,
-    max_iter=300,
-    verbose=False,
-    x_squared_norms=None,
-    tol=1e-4,
-    n_threads=1,
-):
+# HACK: daal4py will fail to import with too recent versions of sklearn because
+# of this missing attribute. Let's pretend it still exists.
+
+if not hasattr(sklearn.neighbors._base, "_check_weights"):
+    warnings.warn(
+        f"The current version of scikit-learn ( =={sklearn.__version__} ) is too "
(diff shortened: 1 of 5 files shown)


fcharras:
Note that while it's an overall improvement over the previous state, it's still hacky (still partially relies on monkey patching for sklearn intelex, and for parameterization of the driver) but I don't think it can get much better.

sklearn_numba_dpex/testing/config.py
```diff
@@ -0,0 +1,14 @@
+from contextlib import contextmanager
+
+
+@contextmanager
+def override_attr_context(obj, **attrs):
+    attrs_before = dict()
+    for attr_name, attr_value in attrs.items():
+        attrs_before[attr_name] = getattr(obj, attr_name)
+        setattr(obj, attr_name, attr_value)
```
jjerphan, line 9:
```suggestion
        attrs_before[attr_name] = getattr(obj, attr_name, default=None)
        setattr(obj, attr_name, attr_value)
```

benchmark/ext_helpers/daal4py.py
```diff
@@ -34,18 +62,12 @@ def kmeans(
...
-#     sample_weight,
-#     centers_init,
-#     max_iter=300,
-#     verbose=False,
-#     x_squared_norms=None,
-#     tol=1e-4,
-#     n_threads=1,
-# ):
+
+# def kmeans_daal4py_kmeans_singlesingle(self, X, sample_weight, centers_init):
```
jjerphan, line 69: Super minor nit.
```suggestion
# def kmeans_daal4py_kmeans_single(self, X, sample_weight, centers_init):
```

benchmark/kmeans.py
```diff
@@ -52,7 +46,7 @@ def __init__(
         self.results = None
         self.skip_slow = skip_slow
 
-    def timeit(self, kmeans_single_lloyd_fn, name, slow=False):
+    def timeit(self, name, engine_provider=None, slow=False):
```
jjerphan, line 49: `engine_provider` must be added to the docstring.

Super minor nitpick: what about renaming `slow` to `is_slow`?

benchmark/ext_helpers/daal4py.py
```diff
@@ -1,28 +1,56 @@
...
-    x_squared_norms=None,
-    tol=1e-4,
-    n_threads=1,
-):
+import sklearn
+import warnings
+
+
+# HACK: daal4py will fail to import with too recent versions of sklearn because
+# of this missing attribute. Let's pretend it still exists.
```
jjerphan, line 6: Fine workaround for me.

jjerphan (approved):
LGTM. Thank you, @fcharras.

sklearn_numba_dpex/testing/config.py
fcharras: In fact not inputing a default is in purpose here, the context manager is called `override` because I'd like it to be used only to override existing attributes, and raising `AttributeError` if trying to override an attribute that does not exist.

(also, it's simpler this way, else to clean properly in the exit we should track attributes that exist and those who don't...)

I'll add a docstring to make this more understandable.

PR #25, opened 2022-09-29
Add support for sample_weight to the KMeans driver

### LINKED ISSUE #22: The KMeans engine does not support non-uniform sample_weight
Trying to use the KMeans engine in `sklearn_numba_dpex` currently results in a `NotSupportedByEngineError` exception.

fcharras:
Fixes #22

This PR also contains some tweaks to conform to all sklearn tests that are not xfailed for `NotSupportedBeEngine` reasons. It now passes successfully 132 tests and xfails 54.

There's only one test that can't pass yet which is the test on relocalization of centroids of empty clusters.

For the time being I propose disabling this test, and remove the `continue-on-error` attribute of the sklearn test pipeline so that the sklearn test module for KMeans becomes a required pass.

I've opened an issue to track progress on the relocalization issue.


fcharras:
The following file (rename its extension from `txt` to  `.html`  and open it in browser, it has a nice interactive menu) show that all the xfailed tests are either related to sparse or to elkan tests:  

[pytest_report.txt](https://github.com/soda-inria/sklearn-numba-dpex/files/9676877/pytest_report.txt)

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -278,14 +322,34 @@ def fused_lloyd_single_step(
...
+
+        # The privatization is more effective when there is a low number of centroid
+        # values (equal to (n_clusters * n_features)) comparatively to the global
+        # number of work items, i.e. when the probability of collision is high. At the
+        # opposite end where the probability of collision is low, privatization might
+        # be detrimental to performance and we might prefer simpler, faster code
+        # with updates directly made into the final array of centroids.
+
+        # The privatization strategy also applies to the updates of the centroid
+        # counts.
```
ogrisel, line 352: The comment blocks above have been duplicated by mistake.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -295,14 +359,14 @@ def fused_lloyd_single_step(
...
-            one,
+            weight if with_sample_weight else one
         )
 
         for feature_idx in range(n_features):
             dpex.atomic.add(
                 new_centroids_t_private_copies,
                 (privatization_idx, feature_idx, min_idx),
-                X_t[feature_idx, sample_idx],
+                X_t[feature_idx, sample_idx] * weight if with_sample_weight else X_t[feature_idx, sample_idx],
```
ogrisel, line 369: Have you run black on this? I am surprised our linter does not complain.

ogrisel: I wonder what is the performance impact of putting the `if with_sample_weight` test withing the `for feature_idx in range(n_features)` loop.

Can you check if putting it outside would not improve the speed?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -426,17 +412,34 @@ def _lloyd(
...
+            #
+            # if pseudo_inertia < best_pseudo_inertia:
+            #     best_pseudo_inertia = pseudo_inertia
+            #     copyto_kernel(centroids_t, best_centroids_t)
+            #
+            # Note that what that is saved as "best centroid" is the array before the
+            # update, to which the pseudo inertia that is computed at this iteration
+            # refers. For this reason, this strategy is not compatible with sklearn
+            # unit tests, that consider new_centroids_t (the array after the update)
+            # to be the best centroids at  each iteration.
```
ogrisel, line 442: nit:

```suggestion
            # to be the best centroids at each iteration.
```

ogrisel: BTW, shall we open a PR to implement this "best centroid" strategy in scikit-learn's Cython k-means? It seems more correct from a mathematical point of view but I am not 100% sure why an update would increase the inertia. Maybe it's just rounding errors, in which case the current code might be fine.

This best centroid check could be useful if we were to implement early stopping on a validation set though.

Maybe @jeremiedbb has an opinion on this.

ogrisel (approved):
LGTM up to following suggestions:

ogrisel:
BTW can you please re-run the benchmark on this branch vs `main` to check that the performance in the case when `sample_weight is None` is not negatively impacted by this PR?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -241,6 +240,9 @@ def _lloyd(
             n_samples,
             n_features,
             n_clusters,
+            with_sample_weight=sample_weight is not None,
+            return_inertia=bool(verbose),
+            compute_exact_inertia=True,
```
jjerphan, line 245: Based on my other comment, I think `compute_exact_inertia` should not be exposed but be defined internally based on the value of `with_sample_weight` and `return_inertia`.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -718,26 +745,35 @@ def _set_dtype(self, X, centers_init):
...
 
-        X, cluster_centers, compute_dtype, output_dtype = self._set_dtype(
-            X, cluster_centers
-        )
+        return X, sample_weight, centers_init, compute_dtype, output_dtype
+
+    def _check_inputs(self, X, sample_weight, cluster_centers):
+
+        if sample_weight is not None and not (sample_weight != sample_weight[0]).any():
+            sample_weight = None
```
jjerphan, line 768: If `sample_weight` is provided as a vector of equal weights, I think we still need to treat it as such because this might be an (I agree weird) intention from the user. In this case, I would change this test to only perform this optimisation in case of unary weights:
```suggestion
        # If sample weights are unary, we can simply ignore them.
        if sample_weight is not None and not (sample_weight != 1).any():
            sample_weight = None
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -777,4 +814,10 @@ def _load_transposed_data_to_device(self, X, cluster_centers):
...
                 f"Expected X_layout to be equal to 'C' or 'F', but got {self.X_layout} ."
             )
-        return X_t, dpctl.tensor.from_numpy(cluster_centers.T, device=self.device)
+        return (
+            X_t,
+            dpctl.tensor.from_numpy(sample_weight, device=self.device)
+            if sample_weight is not None
+            else dpctl.tensor.empty(1, dtype=X.dtype, device=self.device),
+            dpctl.tensor.from_numpy(cluster_centers.T, device=self.device),
+        )
```
jjerphan, line 823: Does this read better?
```suggestion
        sample_weights = (
            dpctl.tensor.from_numpy(sample_weight, device=self.device)
            if sample_weight is not None
            else dpctl.tensor.empty(1, dtype=X.dtype, device=self.device)
        )
        Y_t = dpctl.tensor.from_numpy(cluster_centers.T, device=self.device)
        return X_t, sample_weights, Y_t
```

jjerphan (commented):
Thanks for this extension, @fcharras.

I have several comment including an important one (marked as such).

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
fcharras: My hope (here and for all the other occurences of this strategy in this PR) are that since `with_sample_weight` is a constant (not a direct input to the kernel) the compiler will detect the dead branch and ignore it when compiling, I've read that this is a common optimization optimizers can do, and also I've not seen differences in the benchmarks before/after this PR.

I'll check wether there's a performance hit, but if there is, having the `if with_sample_weight` is still really inconvenient here, we can't use `dpex.func` because there is an atomic op, so we would be left with having to copy/paste the whole block :-/

Another possibility is to set `weight = 1.0` if `with_sample_weight = False` but I would expect that not to be optimized by the compiler and really cost an additional multiply..

sklearn_numba_dpex/kmeans/drivers.py
fcharras: I don't have an opinion on this, frankly I doubt it really has a practical implication and sounds like over-optimization, ideally someone with more knowledge about KMeans could help us decide.

In any case, transitioning to either one or the other strategy is straightforward, and we will stick to whatever strategy conforms to sklearn unit tests.

Maybe the wall of inline text I let in the code gives too much importance to the issue :grin:

sklearn_numba_dpex/kmeans/drivers.py
ogrisel: +1 but with `s/sample_weights/sample_weights/` to be consistent with the existing naming.

sklearn_numba_dpex/kmeans/drivers.py
ogrisel: Actually I think we should not set `sample_weight = None` even when `sample_weight == np.ones(n_samples)`.

This will make it possible to check that the computation with `sample_weight is None` returns the same centroids / labels as with `sample_weight == np.ones(n_samples)` while being faster.

sklearn_numba_dpex/kmeans/drivers.py
ogrisel: I think it's fine to keep the inline comment.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -249,6 +275,24 @@ def fused_lloyd_single_step(
...
+        if with_sample_weight:
+            weight = sample_weight[sample_idx]
+
+        if return_inertia:
+            inertia = min_sample_pseudo_inertia
+            if compute_exact_inertia:
+                inertia = X_l2_norm + (two * min_sample_pseudo_inertia)
+            if with_sample_weight:
+                inertia = inertia * weight
+            per_sample_inertia[sample_idx] = inertia
```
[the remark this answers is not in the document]
fcharras: >  I think we can remove the first branch

The weight is reused when updating centers later in the kernel, that's why the branch is used here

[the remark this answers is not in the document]
fcharras: I think that for the name of the variables there's a confusion between "inertia" and "distance to closest centroid" now with the introduction of "sample_weight" in the mix.

But overall I think the logic is good ? "pseudo_inertia" refers to inertia up to a constant that depends on X and is constant across iterations, I think this still holds even with sample_weight ? (that the constant changes does not matter, basically the constant = the weighted sums of squared l2 norms of X samples)

(NB: in practice it's not used anymore by the driver, I let this option in case we want to return to the previous operation where "best inertia" and "best center" are tracked, should I completely remove it ?)
-

sklearn_numba_dpex/kmeans/drivers.py
fcharras: middle ground: still apply the optimization but multiply the total inertia by the constant just before it's printed / returned (from the user point of view, will have the same effect)

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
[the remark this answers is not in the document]
fcharras: Copying a reply in a past comment about `with_sample_weight` use in general:

> My hope (here and for all the other occurences of this strategy in this PR) are that since `with_sample_weight` is a constant (not a direct input to the kernel) the compiler will detect the dead branch and ignore it when compiling, I've read that this is a common optimization optimizers can do, and also I've not seen differences in the benchmarks before/after this PR.

sklearn_numba_dpex/kmeans/drivers.py
fcharras: > Actually I think we should not set `sample_weight = None` even when `sample_weight == np.ones(n_samples)`.

@ogrisel the issue is that sklearn always transforms a `None` into an array of ones, if we don't force this optimization here, with the current code it will never happen because sklearn force the opposite (and our engine currently inherits from all this preparation)

ogrisel: Ok then. Have you measured the performance impact? If it's not significant, we might was well do as in the original scikit-learn code base and drop the unweighted kernels.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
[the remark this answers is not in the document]
ogrisel: +1 for cleaning up YAGNIs.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
ogrisel: Let's benchmark if `weight = 1.0` is causing a slow-down or not. If not, we can simplify the code.

Related to the discussion in https://github.com/soda-inria/sklearn-numba-dpex/pull/25#discussion_r985523816.

fcharras:
Some benchmark results with and without sample weights (from the benchmark script, which have been updated, + a timeit of the core kernel):
```
# With random sample_weight

Running Sklearn vanilla lloyd with parameters sample_weight=random n_clusters=127 data_shape=(5265120, 14) max_iter=100 tol={tol} ...
Running Sklearn vanilla lloyd ... done in 36.12952249200316

Running daal4py lloyd CPU with parameters sample_weight=random n_clusters=127 data_shape=(5265120, 14) max_iter=100 tol={tol} ...
NotSupportedByEngineError('Non unary sample_weight is not supported by daal4py.')

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) with parameters sample_weight=random n_clusters=127 data_shape=(5265120, 14) max_iter=100 tol={tol} ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) ... done in 13.499520910991123

%timeit fused_lloyd_fixed_window_single_step_kernel(X_t,sample_weight,centroids_t,centroids_half_l2_norm,per_sample_inertia,new_centroids_t_private_copies,cluster_sizes_private_copies)
86.2 ms ± 1.07 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# With sample_weight = None

Running Sklearn vanilla lloyd with parameters sample_weight=None n_clusters=127 data_shape=(5265120, 14) max_iter=100 tol={tol} ...
Running Sklearn vanilla lloyd ... done in 34.9267187469959

Running daal4py lloyd CPU with parameters sample_weight=None n_clusters=127 data_shape=(5265120, 14) max_iter=100 tol={tol} ...
Running daal4py lloyd CPU ... done in 16.513784038004815

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) with parameters sample_weight=None n_clusters=127 data_shape=(5265120, 14) max_iter=100 tol={tol} ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) ... done in 13.348536286997842

%timeit fused_lloyd_fixed_window_single_step_kernel(X_t,sample_weight,centroids_t,centroids_half_l2_norm,per_sample_inertia,new_centroids_t_private_copies,cluster_sizes_private_copies)
86.5 ms ± 1.35 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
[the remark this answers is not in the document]
fcharras: :heavy_check_mark: YAGNI cleant up

fcharras:
Same, with two dimensions:


```
# With random sample_weight

Running Sklearn vanilla lloyd with parameters sample_weight=random n_clusters=127 data_shape=(5265120, 2) max_iter=100 tol={tol} ...
Running Sklearn vanilla lloyd ... done in 24.158396113009076

Running daal4py lloyd CPU with parameters sample_weight=random n_clusters=127 data_shape=(5265120, 2) max_iter=100 tol={tol} ...
NotSupportedByEngineError('Non unary sample_weight is not supported by daal4py.')

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) with parameters sample_weight=random n_clusters=127 data_shape=(5265120, 2) max_iter=100 tol={tol} ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) ... done in 6.525093175994698

%timeit fused_lloyd_fixed_window_single_step_kernel(X_t,sample_weight,centroids_t,centroids_half_l2_norm,per_sample_inertia,new_centroids_t_private_copies,cluster_sizes_private_copies)
47.5 ms ± 1.36 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# With sample_weight = None

Running Sklearn vanilla lloyd with parameters sample_weight=None n_clusters=127 data_shape=(5265120, 2) max_iter=100 tol={tol} ...
Running Sklearn vanilla lloyd ... done in 25.68170678599563

Running daal4py lloyd CPU with parameters sample_weight=None n_clusters=127 data_shape=(5265120, 2) max_iter=100 tol={tol} ...
Running daal4py lloyd CPU ... done in 9.045331621993682

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) with parameters sample_weight=None n_clusters=127 data_shape=(5265120, 2) max_iter=100 tol={tol} ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) ... done in 6.388995380999404

%timeit fused_lloyd_fixed_window_single_step_kernel(X_t,sample_weight,centroids_t,centroids_half_l2_norm,per_sample_inertia,new_centroids_t_private_copies,cluster_sizes_private_copies)
46.7 ms ± 796 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

fcharras:
:shrug: i guess the difference is barely noticeable, let's just use unary `sample_weights` rather than `sample_weight=None`

fcharras:
@jjerphan @ogrisel  At what point do you consider at optimization worth ? If those benchmarks have any significance the improvement was less than 1% (~1% if the input have 2 dimensions, else more in the range of 0.1%). All the small optimizations maybe should be considered if it adds up to something more meaningful in the end?

I merge after having applied your suggestions and dismissed the `with_sample_weight` optim. The optimization commit remains in history, will still possible to resurface it another time.

sklearn_numba_dpex/kmeans/drivers.py
jeremiedbb: Inertia is guaranteed to decrease at each iteration, even when a center is relocated. The only way I could see it increase is due to rounding errors, but then it would only increase by a tiny amount. I can't see the benefit of this "best centroids" strategy.

PR #26, opened 2022-09-30
WIP: relocate centroids of empty clusters

fcharras:


sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -416,6 +501,9 @@ def _lloyd(
                 # ???: verbosity comes at the cost of performance since it triggers
                 # computing exact inertia at each iteration. Shouldn't this be
                 # documented ?
+                # sklearn even goes further and re-computes inertia on updated clusters
+                # rather than old clusters but surely it probably isn't worth the
+                # computational cost ?
```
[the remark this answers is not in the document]
ogrisel: There must be a test in the test suite that checks this, no?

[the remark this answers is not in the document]
fcharras: > There must be a test in the test suite that checks this, no?

For the finalization step, yes and our current behavior is good already, this comment refers only to what is printed when verbosity = True

However the test for the verbosity parameter in sklearn [does not test the printed value](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/cluster/tests/test_k_means.py#L403)

fcharras:
Branch out of WIP !

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -396,11 +423,69 @@ def _lloyd(
             )
 
             nb_empty_clusters_ = dpctl.tensor.asnumpy(nb_empty_clusters)[0]
```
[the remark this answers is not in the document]
fcharras: You're right `dpctl.tensor.asnumpy` was not necessary, operators such as < > works with dpctl tensors. It still need to be converted to python or numpy data later though, I think the jit does not accept dpctl-typed constants,

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -89,45 +94,64 @@ def test_kmeans_same_results(dtype):
...
+        kmeans.fit(X)
+
+    expected_n_iter = 3
+    expected_inertia = 0.25
+    assert_allclose(kmeans.inertia_, expected_inertia)
+    assert kmeans.n_iter_ == expected_n_iter
+
+    # There are two acceptable ways of relocating clusters in this example, the output
+    # depends on how the argpartition strategy break ties. It might not be deterministic
+    # (might depend on thread concurrency) so we accept both outputs.
```
jjerphan, line 118: What if we change this test to simultaneously sort both arrays prior to assertions?

I think this can be done by `argsort`-ing on `expected_centers` to get a permutation that we can apply on both `expected_labels` and `expected_centers`. What do you think of this?

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -89,45 +94,64 @@ def test_kmeans_same_results(dtype):
 
 
 @pytest.mark.parametrize("dtype", [np.float32, np.float64])
-def test_kmeans_fit_empty_clusters(dtype):
+def test_kmeans_relocated_clusters(dtype):
+    """Copied and adapted from sklearn's unit tests"""
```
jjerphan, line 98:
```suggestion
def test_kmeans_relocated_clusters__adapted(dtype):
    """Copied and adapted from sklearn's test_kmeans_relocated_clusters"""
```

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -89,45 +94,64 @@ def test_kmeans_same_results(dtype):
...
-        random_state=random_seed,
-        algorithm="lloyd",
-        n_clusters=n_clusters,
-        init=init,
-        n_init=1,
-        max_iter=1,
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_kmeans_relocated_clusters_2(dtype):
+    """Copied and adapted from sklearn's unit tests"""
```
jjerphan, line 133:
```suggestion
def test_kmeans_relocated_clusters__adapted(dtype):
    """Copied and adapted from sklearn's test_kmeans_relocated_clusters."""
```

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -29,10 +26,18 @@ def _fail_if_no_dtype_support(xfail_fn, dtype):
         )
 
 
+def test_dpnp_implements_argpartition():
+    # Detect if dpnp exposes an argpartition kernel and alert that sklearn-numba-dpex
+    # can be adapted accordingly.
+    assert not hasattr(dpnp, "argpartition"), (
+        "An argpartition function is now available in dpnp, it should now be used in "
+        "place of dpnp.partition in sklearn_numba_dpex.kmeans.drivers ."
+    )
```
jjerphan, line 35: :+1:

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,150 @@
...
+            n_selected_eq_threshold,        # OUT      (1, )
+        ):
+    # fmt: on
+        """This kernel writes in selected_samples_idx the idx of the top n_selected
+        items in distance_to_centroid with highest values.
+
+        threshold is expected to have been pre-computed (by partitioning) such that
+        there are at most (n_selected-1) values that are strictly greater than
+        threshold, and at least n_selected values that are greater or equal than
+        threshold.
```
jjerphan, line 128: I find numerical value being parenthesised unusual.

Would it be better to have them backquoted? E.g.
```suggestion
        threshold is expected to have been pre-computed (by partitioning) such that
        there are at most `n_selected-1` values that are strictly greater than
        threshold, and at least n_selected values that are greater or equal than
        threshold.
```

This comment applies to others docstrings.

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,150 @@
...
+
+    return relocate_empty_clusters[global_size, work_group_size]
+
+
+@lru_cache
+def make_select_samples_far_from_centroid_kernel(
+    threshold, n_selected, n_samples, work_group_size
+):
+    global_size = math.ceil(n_selected / work_group_size) * work_group_size
+    zero = np.int64(0)
```
jjerphan, line 107: From all the scalars being defined here, why is `zero` the only to be a `int64`? Is is for indexing reasons?

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -73,7 +217,7 @@ def make_reduce_centroid_data_kernel(
     global_size = n_work_items_for_clusters * n_features
     zero = dtype(0.0)
     i_one = np.int32(1)
-    i_zero = np.int64(0)
+    l_zero = np.int64(0)
```
jjerphan, line 220: Same comment: why is `l_zero` a `int64` here?

jjerphan (commented):
Thanks @fcharras, here are a few comments.

sklearn_numba_dpex/kmeans/drivers.py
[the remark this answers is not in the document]
fcharras: All things considered this comment will just be removed if https://github.com/soda-inria/sklearn-numba-dpex/pull/33 is accepted, then we reach full compliance with sklearn :lotus_position:

sklearn_numba_dpex/tests/test_kmeans.py
fcharras: I don't think the fact that the results differs only by a permutation is not particularly interesting to use, it only holds for this particular example and would not be true in general.

sklearn_numba_dpex/kmeans/kernels/utils.py
fcharras: I like expliciting the types because I'm afraid that else it could give unpractical types that would be implicitly casted during execution.

Maybe i've found that in `dpex.atomic` calls the index must be given in `int64` ? (else it fails ?) not sure anymore

:shrug: or maybe I should just drop expliciting the dtype of the constant and use plain numbers in the kernels.

sklearn_numba_dpex/kmeans/kernels/utils.py
fcharras: :heavy_check_mark: I apply this but about the others IDK how to detect all parenthesis expression in docstrings :-/

sklearn_numba_dpex/kmeans/kernels/utils.py
jjerphan: I was curious whether `np.int32` is usable instead.

> Maybe i've found that in dpex.atomic calls the index must be given in int64 ? (else it fails ?) not sure anymore

I suspect this this is the main reason.

> or maybe I should just drop expliciting the dtype of the constant and use plain numbers in the kernels.

I also like explicit types. +0 on changing to plain numbers.

sklearn_numba_dpex/kmeans/kernels/utils.py
jjerphan: Ah, I meant parenthesised expressions in docstrings and comments introduced here can be manually be changed to use backquotes. Once again that's a nit.

fcharras:
LGTM ?

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,150 @@
...
+        new_location_X_idx = samples_far_from_center[n_selected_gt_threshold - relocated_idx]
+        new_location_previous_assignment = assignments_idx[new_location_X_idx]
+
+        new_centroid_value = X_t[feature_idx, new_location_X_idx]
+        new_location_weight = sample_weight[new_location_X_idx]
+        X_centroid_addend = new_centroid_value * new_location_weight
+
+        # Cancel the contribution to the updated centroids of the sample that was once
+        # assigned to new_location_previous_assignment but is now assigned to the
+        # cluster of the centroids that relocates to this sample
```
jjerphan, line 78: I find this comment a bit lengthy. Can this rephrased? For instance:

```suggestion
        # Cancel the contribution to new_location_previous_assignment, the 
        # previously updated centroids of the sample.
        # This contribution will be assigned to the centroid of the clusters
        # that relocates to this sample.
```

Note that I turned "the cluster of the centroids" to "the centroid of the clusters".

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,150 @@
...
+            centroids_t,
+            (feature_idx, new_location_previous_assignment),
+            X_centroid_addend
+            )
+
+        # The relocated centroid has only one contribution now, which is the sample
+        # to which it has been relocated to
+        centroids_t[feature_idx, relocated_cluster_idx] = X_centroid_addend
+
+        # Likewise, we update the weights in the clusters
```
jjerphan, line 89: Nit.
```suggestion
        # The relocated centroid has only one contribution now, which is 
        # the one of the sample the cluster has been relocated to.
        centroids_t[feature_idx, relocated_cluster_idx] = X_centroid_addend

        # Likewise, we update the weights in the clusters.
        # This is done once using `feature_idx`'s value.
```

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,150 @@
...
+        if feature_idx == zero:
```
jjerphan, line 90: `feature_idx` is likely an integral whilst `zero` is likely a floating. I do not think this causes problems, but I am just checking.

jjerphan (approved):
LGTM.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -383,21 +401,69 @@ def _lloyd(
...
+                    # centroids already have been computed in the main kernel we only
+                    # need to compute assignments
+                    fixed_window_assignment_kernel(
+                        X_t,
+                        centroids_t,
+                        centroids_half_l2_norm,
+                        assignments_idx,
+                    )
+                else:
+                    # else, compute distances + assignments
```
ogrisel, line 443:
```suggestion
                    # else, compute distances + assignments
                    # Note that we intentionally we pass unit weights instead of
                    # sample_weight so that per_sample_inertia will be updated
                    # to the (unweighted) squared distance to the
                    # nearest centroid.
```

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,150 @@
...
+        distance_to_centroid_ = distance_to_centroid[sample_idx]
+        if distance_to_centroid_ < threshold:
+            return
+
+        if distance_to_centroid_ > threshold:
+            selected_idx = dpex.atomic.add(
+                n_selected_gt_threshold,
+                zero,
+                one
+                )
```
ogrisel, line 155:
```suggestion
        if distance_to_centroid_ > threshold:
            selected_idx = dpex.atomic.add(
                n_selected_gt_threshold,
                zero,
                one,
            )
```

sklearn_numba_dpex/kmeans/kernels/utils.py
fcharras: I'm not surprised that it compiles, it seems like in numba_dpex built in functions have as much implementations as there are possible combinations of input types, if an implementation is not available for a given combination of types the compilation fail with an explicit error message that lists all possible implementations and explains what is the combination of types was rejected. I'm more worried of implicit casting within those implementations.

sklearn_numba_dpex/kmeans/kernels/utils.py
fcharras: caught a few more

fcharras:
The latest commits should address all remaining suggestions, the unit test of the argpartition kernel revealed an obvious error + a less obvious one related to numerical stability that I think is solved now.

.github/workflows/run_tests.yaml
```diff
@@ -36,4 +36,14 @@ jobs:
       # upstream so it doesn't have to rely on order in case of equality.
       # Until then this test is disabled but similar tests are internalized in
       # sklearn_numba_dpex.
+      
+      # TODO: test_kmeans_relocated_clusters on the sklearn side relies on the 
+      # arbitrary tie-breaking behavior of the np.argpartition function. The 
+      # within-partition order should be considered an implementation detail and 
+      # sklearn_numba_depx own kernels for argpartition do not attempt to replicate the
```
ogrisel, line 34: The suggestion feature does not show-up on this line but there is a typo:

`sklearn_numba_depx` => `sklearn_numba_dpex`

.github/workflows/run_tests.yaml
```diff
@@ -28,8 +28,13 @@ jobs:
...
       # TODO: run those tests in a separate pipeline
     - name: Run sklearn test suites with sklearn_numba_dpex engines
-      # TODO: enable the test for cluster relocalization when it's properly supported
-      # by the KMeans engine in sklearn_numba_dpex. More information about the
-      # progress toward support is available on the issue tracker at:
-      # https://github.com/soda-inria/sklearn-numba-dpex/issues/24
+      # TODO: test_kmeans_relocated_clusters on the sklearn side relies on the 
+      # arbitrary tie-breaking behavior of the np.argpartition function. The 
+      # within-partition order should be considered an implementation detail and 
+      # sklearn_numba_depx own kernels for argpartition do not attempt to replicate the
```
ogrisel, line 34:
```suggestion
      # sklearn_numba_dpex own kernels for argpartition do not attempt to replicate the 
```

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -176,3 +177,113 @@ def test_inertia(dtype):
...
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_select_samples_far_from_centroid_kernel(dtype):
+    _fail_if_no_dtype_support(pytest.xfail, dtype)
+
+    # The following array, when sorted, reads:
+    # [-1.0, 9.0, 9.0, 20.22, 20.22, 20.22, 20.22, 23.0, 23.0, 30.0]
+    # So we expect 3 values above threshold: 23.0, 23.0, 30.0
+    # and 2 values equal to threshold.
```
ogrisel, line 222:
```suggestion
    # Assuming n_selected == 5, the threshold is 20.22 and we expect 3 values
    # above threshold (23.0, 23.0, 30.0) and 2 values equal to threshold.
```

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -176,3 +177,113 @@ def test_inertia(dtype):
...
+    n_selected = 5
+    n_samples = len(distance_to_centroid)
+    work_group_size = 4
+    select_samples_far_from_centroid_kernel = (
+        make_select_samples_far_from_centroid_kernel(
+            n_selected, n_samples, work_group_size
+        )
+    )
+
+    # NB: the values used to initialize the output array does not matter, 100 is chosen
```
ogrisel, line 258:
```suggestion
    # NB: the values used to initialize the output array do not matter, 100 is chosen
```

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -176,3 +177,113 @@ def test_inertia(dtype):
...
+    # It is only used as an intermediary variable in the kernel and is not used
+    # otherwise
+    n_selected_eq_threshold = int(n_selected_eq_threshold[0] - 1)
+    n_selected_gt_threshold = int(n_selected_gt_threshold[0])
+    selected_samples_idx = dpctl.tensor.asnumpy(selected_samples_idx)
+
+    # NB: the exact number of selected values equal to the threshold and the
+    # corresponding selected indexes in the input array can change depending on
+    # concurrency. We only check conditions for success, those does not depend on the
+    # variability.
```
ogrisel, line 282:
```suggestion
    # concurrency. We only check conditions for success that do not depend on the
    # concurrency induced non-determinism.
```

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -176,3 +177,113 @@ def test_inertia(dtype):
...
+    assert n_selected_gt_threshold == 3
+    assert n_selected_eq_threshold >= 2
+    assert set(selected_samples_idx[:3]) == {3, 7, 9}
+    assert set(selected_samples_idx[-n_selected_gt_threshold:]).issubset({0, 1, 4, 6})
```
ogrisel, line 286: If I am not mistaken, the second assertion should use `n_selected_gt_threshold` instead of `-n_selected_eq_threshold`.

```suggestion
    expected_gt_threshold_indices = {3, 7, 9}
    actual_gt_threshold_indices = set(
        selected_samples_idx[:n_selected_gt_threshold]
    )  # stored at the beginning
    assert actual_gt_threshold_indices == expected_gt_threshold_indices
 
    expected_eq_threshold_indices = {0, 1, 4, 6}
    actual_eq_threshold_indices = set(
        selected_samples_idx[-n_selected_eq_threshold:]
    )  # stored at the end
    assert actual_eq_threshold_indices.issubset(expected_eq_threshold_indices)
```

not sure about by manual code formatting

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,156 @@
...
+    max_n_selected_eq_threshold = np.int32(n_selected + 1)
+
+    @dpex.kernel
+    # fmt: off
+    def select_samples_far_from_centroid(
+            distance_to_centroid,           # IN       (n_samples,)
+            threshold,                      # IN       (n_samples,)
+            selected_samples_idx,           # OUT      (n_samples,)
+            n_selected_gt_threshold,        # OUT      (1, )
+            n_selected_eq_threshold,        # OUT      (1, )
```
ogrisel, line 124: consistency nit:

```suggestion
            n_selected_gt_threshold,        # OUT      (1,)
            n_selected_eq_threshold,        # OUT      (1,)
```

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -25,6 +25,156 @@
...
+
+        threshold is expected to have been pre-computed (by partitioning) such that
+        there are at most `n_selected-1` values that are strictly greater than
+        threshold, and at least n_selected values that are greater or equal than
+        threshold.
+
+        Because the exact number of values strictly equal to the threshold is not known
+        and that the goal is to select the top n_selected greater items above threshold,
+        we write indices of values strictly greater than threshold at the beginning of
+        the array, and indices of values equal to threshold at the end of the array.
```
ogrisel, line 138:
```suggestion
        the array, and indices of values equal to threshold at the end of the
        selected_samples_idx array.
```

ogrisel (approved):
LGTM!

fcharras:
Will merge when CI green :rocket:

PR #32, opened 2022-10-13
Downgrade dpnp from 0.10.2 to 0.10.1 as it causes segfaults

fcharras:


fcharras:
Ty @jjerphan :grin: 

FYI I also opened https://github.com/IntelPython/dpnp/issues/1199

Issue #28, opened 2022-10-07
Rounding errors when computing inertia in KMeans


fcharras:
This issue to track a point brought by @jeremiedbb regarding computing inertia: computing (x-y)2 is more numerically stable than computing x2 + 2*(-xy + y2/2) and where inertia is printed or returned to the user, the numerical stability should be preferred. Our `get_inertia` and `get_labels_and_inertia` kernels should be amended to use (x-y)2

jjerphan:
The current implementation is not the stablest indeed, but should be the most computational intensive one.

It the use of this decomposition currently causing problem for the implementation?

jeremiedbb:
Here's an example where it can fail hard
```py
import numpy as np
from sklearn.cluster import KMeans

X = np.array([[-1.0001],[-0.9999],[0.9999],[1.0001]], dtype=np.float32)
km = KMeans(n_clusters=2, n_init=1, algorithm='full')
km.fit(X)

# The centroids are correct: it's fine to use the x2 + 2*(-xy + y2/2) formula to find the labels
km.cluster_centers_
>>> [[ 1.]
     [-1.]]

# using the x2 + 2*(-xy + y2/2)  [note that this is not from the actual sklearn KMeans :) ]
km.inertia_
>>> 0.0

# correct inertia
((X - km.cluster_centers_[km.labels_])**2).sum()
>>> 4.0013276e-08
```

fcharras:
TY for the reproducing example !

`4e-08` error doesn't sound too bad for float32 compute ? let's keep this issue open and maybe come back at it in later stages.

I think the performance impact of returning the accurate inertia in score or fit method will be unnoticeable, I'm less sure about using that when `verbose=True`

fcharras:
OK. I'll issue a PR to fix this. A side point is that fusing computations of inertia and labels is counter productive, that explains the choices in sklearn that we didn't understand.

PR #33, opened 2022-10-14
Always use numerically stable expressions when computing inertia

fcharras:
Fixes https://github.com/soda-inria/sklearn-numba-dpex/issues/28


fcharras:
I think the performance cost to having numerically stable expression is harmless because (except when verbose=True) it does not interfere with most of the compute that's to say the lloyd steps.

And this change of strategy actually makes code refactoring easier, with this change the codebase is magically 400 lines lighter.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -535,7 +569,13 @@ def _get_labels(
             dtype=compute_dtype,
         )
 
-        X_t, _, centroids_t = self._load_transposed_data_to_device(X, None, centers)
+        X_t, sample_weight, centroids_t = self._load_transposed_data_to_device(
+            X,
+            None
+            if (sample_weight is _IgnoreSampleWeight) and not with_inertia
+            else sample_weight,
```
ogrisel, line 576: For readability I would rather use an extra local variable instead of using a multi-line ternary operator.

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -35,6 +35,7 @@ def make_label_assignment_fixed_window_kernel(
     ) = _make_initialize_window_kernel_funcs(
         n_clusters,
         n_features,
+        True,
```
ogrisel, line 38: Maybe move this boolean flag as the last argument to make it possible to pass it as a named argument to improve readability.

sklearn_numba_dpex/kmeans/kernels/_base_kmeans_kernel_funcs.py
```diff
@@ -6,6 +6,7 @@
 def _make_initialize_window_kernel_funcs(
     n_clusters,
     n_features,
+    load_centroid_window_half_l2_norm,
```
ogrisel, line 9: The position of the word "window" makes it hard to understand the meaning of this flag.

I think this should be renamed to `initialize_centroid_half_l2_norm_window` or even `initialize_window_of_centroids_half_l2_norms`.

Also, let's move this param as the last param to make it possible to pass it as a named param to improve readability of calls to this function.

sklearn_numba_dpex/kmeans/kernels/_base_kmeans_kernel_funcs.py
```diff
@@ -84,92 +59,89 @@ def _load_window_of_centroids_and_features(
...
-            sample_idx,          # PARAM
-            first_feature_idx,   # PARAM
-            X_t,                 # IN
-            centroids_window,    # IN
-            dot_products,        # OUT
+        def _initialize_window_of_centroids(
+            local_work_id,                  # PARAM
+            first_centroid_idx,             # PARAM
+            centroids_half_l2_norm,         # IN
+            centroids_window_half_l2_norm,  # OUT
```
ogrisel, line 80: Similar comment about the naming of this variable:

```suggestion
            centroids_half_l2_norm_window,  # OUT
```

ogrisel (commented):
In general LGTM. I had a bit of trouble understanding the change at first, maybe the following suggestions can help improve readability a bit.

I also did some benchmarks on my local machine and I confirm that performance is never significantly degraded by the PR and actually improves a bit for the largest workgroup size multipliers.

ogrisel:
> to compare main's and this PR's branch's implementations' performance using [asv continuous](https://asv.readthedocs.io/en/stable/commands.html#asv-continuous) doable?

I think the ad-hoc benchmark script (`benchmark/kmeans.py`) in this repo is enough and is anyway useful as it allows for:

- comparing with vanilla KMeans;
- comparing with `scikit-learn-intelex`;
- explore the impact of the workgroup size parameter.

Granted ASV also measures the standard deviation and automatically computes a report with the speed-up between the 2 branches which is quick to copy paste in such PRs.

sklearn_numba_dpex/tests/test_kmeans.py
```diff
@@ -131,3 +134,45 @@ def test_kmeans_fit_empty_clusters(dtype):
         warnings.simplefilter("error")
         with config_context(engine_provider="sklearn_numba_dpex"):
             kmeans_without_empty_cluster.fit(X)
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_euclidean_distance__adapted(dtype):
```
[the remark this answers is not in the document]
fcharras: I added the suffix after @jjerphan suggestion, reviewers having contradictory suggestions :wink: 

I'll side with removing the suffix to break the balance.

fcharras:
TY for benchmarking @ogrisel  (and the commit that fix tol printing) 

I've pushed a commit that addresses the readability suggestions.  

I'll run a quick benchmark to see the impact of verbosity.

ogrisel (approved):
Assuming the benchmarks are still good with the last batch of changes, LGMT!

fcharras:
Performance still OK.

When `verbosity=True` I find an overhead of 5-10% for both vanilla and ours with low dim and k (resp. 15 and 30).  ( @jeremiedbb )

PR #37, opened 2022-10-25
FEA Implement kmeans++ with numba_dpex kernels

fcharras:
Implement kmeans++

Rely on merging of https://github.com/soda-inria/sklearn-numba-dpex/pull/45 and https://github.com/soda-inria/sklearn-numba-dpex/pull/46 beforehand


fcharras:
Out of WIP !

benchmark/kmeans.py
```diff
@@ -196,7 +198,9 @@ def benchmark_data_initialization(
         # X = np.hstack([X for _ in range(5)])
         # X = np.vstack([X for _ in range(20)])
         rng = default_rng(random_state)
-        init = np.array(rng.choice(X, n_clusters, replace=False), dtype=np.float32)
+        global init
+        if init == "random":
+            init = np.array(rng.choice(X, n_clusters, replace=False), dtype=np.float32)
```
jjerphan, line 203: Would it be better not to overwrite `init` and remove the `global` keyword? E.g. with potentially other changes:
```suggestion
        if init == "random":
            centroids_init = np.array(rng.choice(X, n_clusters, replace=False), dtype=np.float32)
```

benchmark/kmeans.py
```diff
@@ -230,7 +234,8 @@ def benchmark_data_initialization(
         skdpex_kmeans_engine_module, KMeansEngine=DAAL4PYEngine
     ), sklearnex_config_context(target_offload="cpu"):
         kmeans_timer.timeit(
-            name="daal4py lloyd CPU", engine_provider="sklearn_numba_dpex"
+            name="daal4py lloyd CPU",
+            engine_provider="sklearn_numba_dpex",
```
jjerphan, line 238: I would revert this change as it is not a semantic one. WDYT?
```suggestion
            name="daal4py lloyd CPU", engine_provider="sklearn_numba_dpex"
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
         )
         return euclidean_distances_t.T, output_dtype
 
+    def kmeans_plusplus(self, X, sample_weight, n_clusters, random_state):
+        centers_t, center_indices, output_dtype = self._kmeans_plusplus(
+            X, sample_weight, n_clusters, random_state
+        )
+        return dpt.asnumpy(centers_t).astype(output_dtype, copy=False), dpt.asnumpy(
+            center_indices
+        )
```
jjerphan, line 802:
```suggestion
        centers_t = dpt.asnumpy(centers_t).astype(output_dtype, copy=False)
        center_indices = dpt.asnumpy(center_indices)

        return centers_t, center_indices
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
...
+
+    def _kmeans_plusplus(self, X, sample_weight, n_clusters, random_state):
+        if X.shape[0] < n_clusters:
+            raise ValueError(
+                f"n_samples={X.shape[0]} should be >= n_clusters={n_clusters}."
+            )
+
+        # NB: the implementation differs from sklearn implementation with regards to
+        # sample_weight, which is ignored in sklearn, but used here.
+        # Check that this is correct ?
```
jjerphan, line 812:
```suggestion
        # NB: the implementation differs from sklearn implementation with regards to
        # sample_weight, which is ignored in sklearn, but used here.
        # TODO: check that this implementation is correct when samples
        # weights aren't uniform.
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
...
+            _,
+            compute_dtype,
+            output_dtype,
+            work_group_size,
+        ) = self._check_inputs(X, sample_weight, cluster_centers=None)
+
+        n_samples, n_features = X.shape
+
+        # heuristic taken from sklearn implementation v<1.2
+        n_local_trials = 2 + int(np.log(n_clusters))
```
jjerphan, line 825: I think this might be read as implicitly implying that the heuristic has been changed in 1.2.

Is this clearer?
```suggestion
        # Same retrial heuristic as scikit-learn (at least until <1.2)
        n_local_trials = 2 + int(np.log(n_clusters))
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
...
+
+        # TODO: this block is also written in common.kernel.random, factorize ?
```
jjerphan, line 827: To me, this can be done as part of another subsequent PR.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -825,8 +1030,9 @@ def _set_dtype(self, X, sample_weight, centers_init):
             # The other cases would not trigger any additional memory copies.
             X = X.astype(compute_dtype)
 
-        centers_init_dtype = centers_init.dtype
-        if centers_init.dtype != compute_dtype:
+        if centers_init is not None and (
+            (centers_init_dtype := centers_init.dtype) != compute_dtype
```
jjerphan, line 1034: Nice use of the walrus operator! :)
Surprisingly enough, that's the first time I get to see it in a review!

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -0,0 +1,224 @@
...
+    dtype,
+):
+
+    rand_uniform_kernel_func = make_rand_uniform_kernel_func(np.dtype(dtype))
+
+    zero_idx = np.int64(0)
+    one_incr = np.int32(1)
+    one_decr = np.int32(-1)
+    zero_init = dtype(0.0)
+    max_candidate_id = np.int32(n_samples - 1)
```
jjerphan, line 78: I like those locals! :100:

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @dpex.kernel
+    # fmt: off
+    def sample_center_candidates(
+        closest_dist_sq,          # IN             (n_features, n_samples)
+        total_potential,          # IN             (1,)
```
jjerphan, line 84: Naive question: what is the meaning of "potential" here? Does this describe inertia?

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -0,0 +1,224 @@
...
+                first_feature_idx += centroids_window_height
+
+            if sample_idx < n_samples:
+                sample_weight_ = sample_weight[sample_idx]
+                closest_dist_sq_ = closest_dist_sq[sample_idx]
+                for i in range(window_n_centroids):
+                    centroid_idx = first_centroid_idx + i
+                    if centroid_idx < n_candidates:
+                        sq_distance_i = min(
+                            sq_distances[i] * sample_weight_,
```
jjerphan, line 215: Same comment regarding squaring the sample weights.

jjerphan (commented):
Thanks for bringing "k-means++" in, @fcharras!

Here is a first yet quite exhaustive review.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -20,6 +23,9 @@
 from sklearn_numba_dpex.testing.config import float_dtype_params
 
 
+driver = KMeansDriver()
```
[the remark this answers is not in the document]
fcharras: no

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
...
+                X_t,
+                sample_weight,
+                candidate_ids,
+                closest_dist_sq,
+                distance_to_candidates_t,
+            )
+
+            candidate_potentials = reduce_potential_2d_kernel(distance_to_candidates_t)[
+                :, 0
+            ]
```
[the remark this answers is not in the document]
fcharras: probably a view but both are OK regarding re-using in future kernels

sklearn_numba_dpex/kmeans/drivers.py
fcharras: I'm a big fan

You don't see it a lot probably because a lot of software still support python 3.7 =)

Also many people dislike because of readability

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -0,0 +1,224 @@
...
+            return
+
+        starting_center_id_ = center_indices[zero_idx]
+
+        sq_distance = zero_init
+        for feature_idx in range(n_features):
+            diff = X_t[feature_idx, sample_idx] - X_t[feature_idx, starting_center_id_]
+            sq_distance += diff * diff
+
+        sq_distance *= sample_weight[sample_idx]
```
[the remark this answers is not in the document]
fcharras: I don't have a source for that, I'd say we shouldn't so that a point that is included two times in the dataset has the same weight that a point with sample_weight = 2.

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
fcharras: I copied vocabulary from sklearn implementation. I don't know why it's called potential.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
...
+                    closest_dist_sq,
+                    total_potential,
+                    random_state,
+                    candidate_ids,
+                )
+
+            # XXX: at the cost of one additional pass on data, we could avoid storing
+            # entirely distance_to_candidates_t in memory, and save
+            # `dtype.nbytes * n_local_trials * n_sample` bytes in memory.
+            # Which is better ?
```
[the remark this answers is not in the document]
fcharras: Just by counting operations, with an additional pass the loop could be about `(1/n_local_trials)` slower with `n_local_trials` taking values up to ~20, so about 5 or 10% slower.

I think it's an acceptable cost compared to the space that is saved, but ultimately it depends on what people value most and what resources are most available.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,202 @@ def _get_euclidean_distances(self, X, Y):
...
+            ]
+            best_candidate = select_best_candidate_kernel(candidate_potentials)[0]
+
+            total_potential = candidate_potentials[
+                best_candidate : (best_candidate + 1)
+            ]
+            closest_dist_sq = distance_to_candidates_t[best_candidate]
+            best_candidate = candidate_ids[best_candidate]
+            centers_t[:, c] = X_t[:, best_candidate]
+            center_indices[c] = best_candidate
```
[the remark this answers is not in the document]
fcharras: Not sure what you means here, in successive iterations `c` and `candidate_ids` are different, and we can expect that once a center has been set, the corresponding points can't be sampled in further iterations because their potential is set to 0.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -277,3 +283,52 @@ def test_select_samples_far_from_centroid_kernel(dtype):
...
+    """Test adapted from sklearn's test_kmeans_plusplus_output"""
+    # Check for the correct number of seeds and all positive values
+    data = sklearn_test_data.astype(dtype)
+
+    centers, indices = driver.kmeans_plusplus(
+        X=data, sample_weight=None, n_clusters=sklearn_test_n_clusters, random_state=42
+    )
+
+    # Check there are the correct number of indices and that all indices are
+    # positive and within the number of samples
```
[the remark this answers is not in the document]
fcharras: we have `indices.shape[0] == sklearn_test_n_clusters` so the number of unique values is at most `sklearn_test_n_clusters` ?

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -277,3 +283,52 @@ def test_select_samples_far_from_centroid_kernel(dtype):
...
+    assert indices.shape[0] == sklearn_test_n_clusters
+    assert (indices >= 0).all()
+    assert (indices <= data.shape[0]).all()
+
+    # Check for the correct number of seeds and that they are bound by the data
```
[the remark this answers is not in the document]
fcharras: We can test that it's either float32 or float64, but I think that more than that is a bit more that what we are set for, since the dtype can change depending on the device...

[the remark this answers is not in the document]
jjerphan: OK, then for now, I am for not asserting dtype invariance and for not casting arrays to other dtype if not strictly required.

sklearn_numba_dpex/kmeans/drivers.py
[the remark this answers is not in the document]
jjerphan: OK, to me this is a bit important for this implementation but really not urgent.

```suggestion
            # XXX: at the cost of one additional pass on data, we could avoid storing
            # entirely distance_to_candidates_t in memory, and save
            # `dtype.nbytes * n_local_trials * n_sample` bytes in memory.
            # TODO: access if this alternative would be more adapted for some regimes.  
```

WDYT?

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
jjerphan: OK, then this looks fine to me.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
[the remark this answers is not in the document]
fcharras: I'm having trouble understanding what you're refering too with so much negatives :grin: do you suggest to revert the change I just made to this test, or change it, or change the behavior of the estimator, or none of those ?

sklearn_numba_dpex/kmeans/drivers.py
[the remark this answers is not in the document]
fcharras: The current version is simpler and already satisfying (same than sklearn) so I like better the idea of merging like this and letting this TODO in comment.

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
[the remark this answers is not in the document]
fcharras: WDYM about lloyd and elkan ? I think weights in lloyd / elkan should only be used in the update step, not when computing closest centroids, and it seems that our implementation and sklearn's agree on that ?

But I think the meaning of weights in kmeans++ is a bit different and there it makes sense to use it to weight distances,

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,211 @@ def _get_euclidean_distances(self, X, Y):
...
+
+            candidate_potentials = reduce_potential_2d_kernel(distance_to_candidates_t)[
+                :, 0
+            ]
+            best_candidate = select_best_candidate_kernel(candidate_potentials)[0]
+
+            total_potential = candidate_potentials[
+                best_candidate : (best_candidate + 1)
+            ]
+            closest_dist_sq = distance_to_candidates_t[best_candidate]
```
fcharras, line 994:
```suggestion
            closest_dist_sq = distance_to_candidates_t[best_candidate, :]
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,211 @@ def _get_euclidean_distances(self, X, Y):
...
+            best_candidate = candidate_ids[best_candidate]
+            centers_t[:, c] = X_t[:, best_candidate]
+            center_indices[c] = best_candidate
```
jjerphan, line 997: To use two different variables for those two index while preserving the logic.
```suggestion
            center_index = candidate_ids[best_candidate]
            centers_t[:, c] = X_t[:, center_index]
            center_indices[c] = center_index
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,211 @@ def _get_euclidean_distances(self, X, Y):
...
+
+        # Pick the remaining n_clusters-1 points
+        for c in range(1, n_clusters):
+            # NB: this step consists in highly sequential instructions that can only
+            # be parallelized in n_local_trials threads, it's up to 50x faster to run it
+            # on CPU, and becomes the bottleneck if ran on GPU. Depending on hardware
+            # and weight of data transfers, are there cases where it would be more
+            # efficient to keep it on GPU ?
+            # ???: would it be bad to sample the sample weight once outside the loop and
+            # reuse it at each iteration ?
```
jjerphan, line 959:
```suggestion
            # reuse it at each iteration ?
            
            # Sample indices of candidates using a empirical cumulative density
            # functions built using samples potentials and squared distances
            # each sample's closest centroids. 
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,211 @@ def _get_euclidean_distances(self, X, Y):
...
+            else:
+                sample_center_candidates_kernel(
+                    closest_dist_sq, total_potential, random_state, candidate_ids
+                )
+
+            # XXX: at the cost of one additional pass on data, we could avoid storing
+            # entirely distance_to_candidates_t in memory, and save
+            # `dtype.nbytes * n_local_trials * n_sample` bytes in memory.
+            # Which is better ?
+            kmeansplusplus_single_step_fixed_window_kernel(
```
jjerphan, line 978:
```suggestion
            # For each (sample, candidate)-pair, compute the minimum
            # between their distance and the previous minimum.
            kmeansplusplus_single_step_fixed_window_kernel(
```

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -45,8 +42,27 @@ def prepare_fit(self, X, y=None, sample_weight=None):
                 f"The sklearn_nunmba_dpex engine for KMeans only support the Lloyd algorithm, {algorithm} is not supported."
             )
 
+        self.sample_weight = sample_weight
+
         return super().prepare_fit(X, y, sample_weight)
 
+    def init_centroids(self, X):
+        if hasattr(self.init, "startswith") and self.init == "k-means++":
```
jjerphan, line 50:
```suggestion
        if instanceof(self.init, str) and self.init == "k-means++":
```

jjerphan (approved):
LGTM up to the last comments made a call between @fcharras and I.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -784,6 +793,211 @@ def _get_euclidean_distances(self, X, Y):
...
+            )
+
+            candidate_potentials = reduce_potential_2d_kernel(distance_to_candidates_t)[
+                :, 0
+            ]
+            best_candidate = select_best_candidate_kernel(candidate_potentials)[0]
+
+            total_potential = candidate_potentials[
+                best_candidate : (best_candidate + 1)
+            ]
```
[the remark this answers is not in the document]
jjerphan: Do you think this comment is worth integrating?

jjerphan (approved):
LGTM. Just one remark regarding a remaining comment.

benchmark/kmeans.py
```diff
@@ -210,7 +209,7 @@ def benchmark_data_initialization(
                 f" {sample_weight} instead."
             )
 
-        return X, sample_weight, init
+        return X, sample_weight, init, n_clusters
 
     kmeans_timer = KMeansLloydTimeit(
```
ogrisel, line 214: I have tried to run the benchmark against this branch and I got:

```python-traceback
Traceback (most recent call last):
  File "/io/code/sklearn-numba-dpex/benchmark/kmeans.py", line 214, in <module>
    kmeans_timer = KMeansLloydTimeit(
  File "/io/code/sklearn-numba-dpex/benchmark/kmeans.py", line 47, in __init__
    ) = data_initialization_fn(**data_initialization_kwargs)
TypeError: benchmark_data_initialization() missing 1 required positional argument: 'init'
```

I did not try to debug yet.

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -0,0 +1,224 @@
...
+    # fmt: off
+    def kmeansplusplus_single_step(
+        X_t,                               # IN READ-ONLY   (n_features, n_samples)
+        sample_weight,                     # IN READ-ONLY   (n_samples,)
+        candidates_ids,                    # IN             (n_candidates,)
+        closest_dist_sq,                   # IN             (n_samples,)
+        euclidean_distances_t,             # OUT            (n_candidates, n_samples)
+    ):
+    # fmt: on
+        sample_idx = dpex.get_global_id(zero_idx)
```
ogrisel, line 160: Since we parallelize on `sample_idx`, I would expect this code to be the fastest when `X_t` has a Fortran layout (contiguous features for a given sample).

But at the moment we raise `ValueError` instead of allocating `X_t` with Fortran layout (that would correspond to C layout for an on-device `X` that does not exist) in `_load_transposed_data_to_device`. But I don't understand why. Both layouts seems to work fine in my setting with dpctl.__version__ == "0.13.0" from today's docker container:

```python
>>> import numpy as np
>>> import dpctl
>>> dpctl.__version__
'0.13.0'
>>> gpu_device = dpctl.SyclDevice("gpu")
<dpctl.SyclDevice [backend_type.level_zero, device_type.gpu,  Intel(R) Iris(R) Graphics 540 [0x1926]] at 0x7ff09594b370>

>>> data_np = np.arange(12).reshape(3, 4)
>>> data_np.T
array([[ 0,  4,  8],
       [ 1,  5,  9],
       [ 2,  6, 10],
       [ 3,  7, 11]])

>>> data_t_c = dpctl.tensor.from_numpy(data_np.T, device=gpu_device)
>>> data_t_c.strides
(3, 1)  # C layout for `data_t_c`
>>> dpctl.tensor.asnumpy(data_t_c)
array([[ 0,  4,  8],
       [ 1,  5,  9],
       [ 2,  6, 10],
       [ 3,  7, 11]])

>>> data_t_f = dpctl.tensor.from_numpy(data_np, device=gpu_device).T
>>> data_t_f.strides
(1, 4)  # F layout for data_t_f
>>> dpctl.tensor.asnumpy(data_t_f)
array([[ 0,  4,  8],
       [ 1,  5,  9],
       [ 2,  6, 10],
       [ 3,  7, 11]])
```

So why don't we work with `X_t` allocated on-device with a Fortran layout?

ogrisel (commented):
Here is a pass of review on the kmeans++ part of the PR.

I have concerns about the way we manage the memory layout. If you prefer, please feel free to open an issue to track this problem and solve it in a follow-up PR. But still I would be interested in your feedback.

My other main concerns would be to add some kind of statistical test on a toy dataset with well separated clusters to check that this implementation lead to inits qualities as good on average as the ones obtained by kmeans++ init of the default engine and significantly better than the random init strategy. See details in comments below.

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -0,0 +1,224 @@
...
+    zero_idx = np.int64(0)
+
+    @dpex.kernel
+    # fmt: off
+    def kmeansplusplus_single_step(
+        X_t,                               # IN READ-ONLY   (n_features, n_samples)
+        sample_weight,                     # IN READ-ONLY   (n_samples,)
+        candidates_ids,                    # IN             (n_candidates,)
+        closest_dist_sq,                   # IN             (n_samples,)
+        euclidean_distances_t,             # OUT            (n_candidates, n_samples)
```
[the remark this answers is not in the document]
fcharras: +1 for `sq_distances_t`, the naming is inspired from sklearn implementation but it's true that `euclidean_distances_t` make sense in sklearn because the full pairwise distance matrix is fully computed, while it's fused with the `min` step in our kernel.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -277,3 +284,57 @@ def test_select_samples_far_from_centroid_kernel(dtype):
...
+    assert centers.shape[0] == n_clusters_sklearn_test
+    assert (centers.max(axis=0) <= X.max(axis=0)).all()
+    assert (centers.min(axis=0) >= X.min(axis=0)).all()
+    # NB: dtype can change depending on the device, so we accept all valid dtypes.
+    assert centers.dtype.type in {np.float32, np.float64}
+
+    # Check that indices correspond to reported centers
+    assert_allclose(X[indices].astype(dtype), centers)
+
+
```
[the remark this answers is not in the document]
fcharras: AFAIK there aren't such tests in sklearn, only basic sanity checks. I can try setting this up.

benchmark/kmeans.py
fcharras: Oops, easy fix fortunately

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
fcharras: `numba_dpex` [does not](https://github.com/IntelPython/numba-dpex/issues/767) support inputs with `F-layout` so it's hard to try our kernels where different layouts, it would require a hack that consists in passing `X` or `X_t` in `C-layout` and inverting the arguments of all `getitem` operations in the kernels depending on wether `X` or `X_t` is passed (perhaps could sound easier if mentionning that `getitem`s could be moved to abstracted kernels that take the layout as instantiation parameter but it would be very verbose for just `getitem`s)

From the information I've gathered before starting the project, our gpu kernels and especially pairwise distance-like kernels should perform better when `X` is `F-layout` so that's why we use `X_t` everywhere.

For the sampling step we could use X in C-Layout if it's more efficient but if changing layout triggers a copy (does it?) it wouldn't be worth it, maybe a compute overhead would be acceptable.

ogrisel: I opened to clarify https://github.com/soda-inria/sklearn-numba-dpex/pull/53 this.

sklearn_numba_dpex/kmeans/drivers.py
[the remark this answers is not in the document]
fcharras: probably missed that comment, will add it in next push

fcharras:
I've pushed the test on quality of initialization, it confirms the validity of our implementation (and sklearn's).

The test runs only 10 different seeds and is already a bit slow, do you think of any suggestions to make it faster ? also it tests exact values of the mean inertia, if you have suggestions of a range that would make the test pass with other seeds while remaining significant it could be more satisfying too.

(also, the results holds with more than 10 seeds and show closer mean inertia for skearn kmeans++ and ours)

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -277,3 +284,119 @@ def test_select_samples_far_from_centroid_kernel(dtype):
...
+            sample_weight=None,
+            n_clusters=n_clusters,
+            random_state=random_state,
+        )
+        scores_engine_kmeans_plusplus.append(
+            _get_score_with_centers(engine_kmeans_plusplus_centers)
+        )
+
+    # Those results confirm that both sklearn KMeans++ and ours have similar quality,
+    # and are both very significantly better than random init.
```
ogrisel, line 336: It's great to see that both kmeans++ are better than random init. However the difference between the 2 kmeans++ is not expected. Have you tried to see if this goes away when increasing the number of iterations from 10 to 100 for instance?

If that's the case please make it explicit in the comment (even if the test still runs 10 iterations to keep it fast to run).

fcharras: Yes, the difference does goes away.

fcharras:
No it fails because I pushed 100 iter by error... sorry, fixing back

fcharras:
About the performances, the changes introduced in https://github.com/soda-inria/sklearn-numba-dpex/pull/49 will also apply to kmeans++.  I'd prefer merging this before adapting https://github.com/soda-inria/sklearn-numba-dpex/pull/49 to include kmeans++ or else, more conflicts :skull:

fcharras:
@ogrisel
> The benchmark script complains that the daal4py engine does not implement kmeans++

[You seem to be right](https://github.com/intel/scikit-learn-intelex/blob/develop/daal4py/sklearn/cluster/_k_means_0_23.py#L86) I erroneously thought that sklearn intelex defaulted back to sklearn if init==kmeans++ but it seems that was wrong.

fcharras:
On my machine it also compares well with `daal4py`

```
Running Sklearn vanilla lloyd ... done in 3.235599868999998

Running daal4py lloyd CPU ... done in 2.575907283000106

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=1) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=1) ... done in 1.9777851419999024

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) ... done in 2.0436579570000504

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=4) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=4) ... done in 2.390581452999868

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=8) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=8) ... done in 2.4388396619997366
```

ogrisel:
Merged!

fcharras:
Awesome! TY for review!!

PR #40, opened 2022-10-27
Use engine-style implementation for daal4py benchmark

fcharras:
The purpose of this PR is to make it easier to write additional methods to the daal4py engine used in benchmarks, in particular the `prepare_fit` i also defined to raise errors for unsupported parameters.

To merge before https://github.com/soda-inria/sklearn-numba-dpex/pull/37 and after https://github.com/soda-inria/sklearn-numba-dpex/pull/39

### DIFF
--- a/benchmark/ext_helpers/daal4py.py
+++ b/benchmark/ext_helpers/daal4py.py
@@ -1,6 +1,5 @@
-import sklearn
 import warnings
-
+import sklearn
 
 # HACK: daal4py will fail to import with too recent versions of sklearn because
 # of this missing attribute. Let's pretend it still exists.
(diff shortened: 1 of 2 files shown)


benchmark/ext_helpers/daal4py.py
```diff
@@ -21,45 +20,63 @@
 else:
     from daal4py.sklearn.cluster._k_means_0_23 import KMeans
 
-
 from sklearn.exceptions import NotSupportedByEngineError
+from sklearn_numba_dpex.kmeans.engine import KMeansEngine
+
+
+# TODO: instead of relying on monkey patching the defau;t engine, find a way to
```
ogrisel, line 27:
```suggestion
# TODO: instead of relying on monkey patching the default engine, find a way to
```

benchmark/ext_helpers/daal4py.py
```diff
@@ -21,45 +20,63 @@
...
+# register a distinct entry point that can load a distinct engine outside of setup.py
+# (impossible ?)
```
ogrisel, line 29: Not sure it's worth it for now. We will reconsider this once we have implemented several engine entrypoints in wip-engines.

ogrisel (approved):

benchmark/ext_helpers/daal4py.py
fcharras: Agree, however it's also an important weakness of the current entry point approach, with the current state of our plugin system users can't just write a new engine in a jupyter notebook and load it into the plugin api, it seems that entry points are _required_ to be declared in the setup file. Maybe we could prioritize investigating this point more, because it might be a deterrent for using `entry_points` at all.

PR #42, opened 2022-10-27
Add a module for rng kernels and kernel funcs

fcharras:
To merge before https://github.com/soda-inria/sklearn-numba-dpex/pull/37 but after https://github.com/soda-inria/sklearn-numba-dpex/pull/40

This implements base functions and kernels for rng that can be used in other kernels.

Not sure about how it should be tested ? I'm thinking that I can only test that the generated values are the same than another [reference implementation](https://bashtage.github.io/randomgen/bit_generators/generated/randomgen.xoroshiro128.Xoroshiro128.random_raw.html#randomgen.xoroshiro128.Xoroshiro128.random_raw) of the same algorithm with some given seeds and add a unit test that hardcode the values, I hope the implementation are exactly identitcal.

### DIFF
--- a/sklearn_numba_dpex/common/random.py
+++ b/sklearn_numba_dpex/common/random.py
@@ -1,0 +1,232 @@
+import warnings
+import random
+import dpctl
+from functools import lru_cache
+from numba import float32, float64, uint32, int64, uint64
+
+import numba_dpex as dpex
+
+import numpy as np
+
+# This code is largely inspired from the numba.cuda.random module and the
+# numba/cuda/random.py where it's defined (v<0.57).
+
+# NB: original code also includes functions for generating normally distributed floats
+# but we don't include it here as long as it's not needed.
+
+zero_idx = int64(0)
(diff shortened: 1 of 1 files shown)


sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,232 @@
...
+import dpctl
+from functools import lru_cache
+from numba import float32, float64, uint32, int64, uint64
+
+import numba_dpex as dpex
+
+import numpy as np
+
+# This code is largely inspired from the numba.cuda.random module and the
+# numba/cuda/random.py where it's defined (v<0.57).
```
jjerphan, line 12: Thank you for indicating this. Can you include a permalink to those implementations?

jjerphan (commented):
Thank you, @fcharras.

Would it make sense to statistically test PRNG kernels for proper pseudo-randomness (e.g. using binning and χ² tests)?

PR #45, opened 2022-10-27
Rng kernels

fcharras:
Same PR than https://github.com/soda-inria/sklearn-numba-dpex/pull/42 that I closed by mistake

This time, with tests

The tests take randomgen implementation as reference, but I've found errors in randomgen while writing the tests, until the errors are confirmed (reported in https://github.com/bashtage/randomgen/issues/321 ) our implementation is corrupted with the same errors for the moment. 

Maybe another reference implementation would be preferable if you have suggestions. I prefer not to use `numba.cuda.random` as reference because it requires a GPU to check it.

### DIFF
--- a/sklearn_numba_dpex/common/random.py
+++ b/sklearn_numba_dpex/common/random.py
@@ -300,7 +300,8 @@ def create_xoroshiro128pp_states(n_states, subsequence_start=0, seed=None, devic
             from_cpu_to_device = True
         except dpctl.SyclDeviceCreationError:
             warnings.warn(
-                "No CPU found, fallbacking random initiatlization to default " "device."
+                "No CPU found, falling back random initiatlization to default "
+                "device."
             )
 
     states = dpt.empty(
--- a/sklearn_numba_dpex/common/random.py
+++ b/sklearn_numba_dpex/common/random.py
@@ -1,36 +1,44 @@
 import warnings
 import random
-import dpctl
 from functools import lru_cache
+
 from numba import float32, float64, uint32, int64, uint64
-
-import numba_dpex as dpex
-
 import numpy as np
 
-# This code is largely inspired from the numba.cuda.random module and the
-# numba/cuda/random.py where it's defined (v<0.57).
+import dpctl
+import dpctl.tensor as dpt
+import numba_dpex as dpex
 
-# NB: original code also includes functions for generating normally distributed floats
-# but we don't include it here as long as it's not needed.
+# This code is largely inspired from the numba.cuda.random module and the
+# numba/cuda/random.py where it's defined (v<0.57), and by the implementation of the
+# same algorithm in the package `randomgen`.
+
+# numba.cuda.random :https://github.com/numba/numba/blob/0.56.3/numba/cuda/random.py
+# randomgen: https://github.com/bashtage/randomgen/blob/v1.23.1/randomgen/xoroshiro128.pyx
+
+# NB1: we implement xoroshiro128++ rather than just xoroshiro128+, which is preferred.
+# Reference resource about PRNG: https://prng.di.unimi.it/
+
+# NB2: original numba.cuda.random code also includes functions for generating normally
+# distributed floats but we don't include it here as long as it's not needed.
 
 zero_idx = int64(0)
 one_idx = int64(1)
 
 
-def get_randint(states):
-    result = dpctl.tensor.empty(sh=(1,), dtype=np.uint64, device=states.device)
-    make_randint_kernel()(states, result)
+def get_random_raw(states):
+    result = dpt.empty(sh=(1,), dtype=np.uint64, device=states.device)
+    make_random_raw_kernel()(states, result)
     return result
 
 
 @lru_cache
-def make_randint_kernel():
+def make_random_raw_kernel():
     @dpex.kernel
-    def _get_randint_kernel(states, result):
-        result[zero_idx] = _xoroshiro128p_next(states, zero_idx)
+    def _get_random_raw_kernel(states, result):
+        result[zero_idx] = _xoroshiro128pp_next(states, zero_idx)
 
-    return _get_randint_kernel[1, 1]
+    return _get_random_raw_kernel[1, 1]
 
 
 @lru_cache
@@ -74,13 +82,13 @@ def make_rand_uniform_kernel_func(dtype):
         )
 
     @dpex.func
-    def xoroshiro128p_uniform(states, index):
-        return uint64_to_unit_float(_xoroshiro128p_next(states, index))
+    def xoroshiro128pp_uniform(states, index):
+        return uint64_to_unit_float(_xoroshiro128pp_next(states, index))
 
-    return xoroshiro128p_uniform
+    return xoroshiro128pp_uniform
 
 
-def create_xoroshiro128p_states(n_states, seed=None, subsequence_start=0, device=None):
+def create_xoroshiro128pp_states(n_states, seed=None, subsequence_start=0, device=None):
     """Returns a new device array initialized for n random number generators.
 
     This initializes the RNG states so that each state in the array corresponds
@@ -111,34 +119,28 @@ def create_xoroshiro128p_states(n_states, seed=None, subsequence_start=0, device
     seed = uint64(seed)
     n_states = int64(n_states)
 
-    init_const_1 = uint64(0x9E3779B97F4A7C15)
-    init_const_2 = uint64(0xBF58476D1CE4E5B9)
-    init_const_3 = uint64(0x94D049BB133111EB)
-    init_rshift_1 = uint32(30)
-    init_rshift_2 = uint32(27)
-    init_rshift_3 = uint32(31)
+    init_const_1 = np.uint64(0)
 
     @dpex.kernel
-    def init_xoroshiro128p_states(states):
+    def init_xoroshiro128pp_states(states):
         """
         Use SplitMix64 to generate an xoroshiro128p state from 64-bit seed.
 
         This ensures that manually set small seeds don't result in a predictable
         initial sequence from the random number generator.
         """
-        if n_states <= one_idx:
+        if n_states < one_idx:
             return
 
-        z = seed + init_const_1
-        z = (z ^ (z >> init_rshift_1)) * init_const_2
-        z = (z ^ (z >> init_rshift_2)) * init_const_3
-        z = z ^ (z >> init_rshift_3)
-        states[zero_idx, zero_idx] = z
-        states[zero_idx, one_idx] = z
+        splitmix64_state = init_const_1 ^ seed
+        splitmix64_state, states[zero_idx, zero_idx] = _splitmix64_next(
+            splitmix64_state
+        )
+        _, states[zero_idx, one_idx] = _splitmix64_next(splitmix64_state)
 
         # advance to starting subsequence number
         for _ in range(subsequence_start):
-            _xoroshiro128p_jump(states, zero_idx)
+            _xoroshiro128pp_jump(states, zero_idx)
 
         # populate the rest of the array
         for idx in range(one_idx, n_states):
@@ -146,17 +148,63 @@ def create_xoroshiro128p_states(n_states, seed=None, subsequence_start=0, device
             states[idx, zero_idx] = states[idx - one_idx, zero_idx]
             states[idx, one_idx] = states[idx - one_idx, one_idx]
             # and jump forward 2**64 steps
-            _xoroshiro128p_jump(states, idx)
+            _xoroshiro128pp_jump(states, idx)
 
-    jump_const_1 = uint64(0xBEAC0467EBA5FACB)
-    jump_const_2 = uint64(0xD86B048B86AA9922)
+    splitmix64_const_1 = uint64(0x9E3779B97F4A7C15)
+    splitmix64_const_2 = uint64(0xBF58476D1CE4E5B9)
+    splitmix64_const_3 = uint64(0x94D049BB133111EB)
+    splitmix64_rshift_1 = uint32(30)
+    splitmix64_rshift_2 = uint32(27)
+    splitmix64_rshift_3 = uint32(31)
+
+    @dpex.func
+    def _splitmix64_next(state):
+        new_state = z = state + splitmix64_const_1
+        z = (z ^ (z >> splitmix64_rshift_1)) * splitmix64_const_2
+        z = (z ^ (z >> splitmix64_rshift_2)) * splitmix64_const_3
+        return new_state, z ^ (z >> splitmix64_rshift_3)
+
+    # TODO: apparently the following implementation should be used in place of the code
+    # that follows afterward. Waiting for randomgen package to confirm that the
+    # implementation they use contains mistakes.
+    # See https://github.com/bashtage/randomgen/issues/321
+
+    # jump_const_1 = uint64(0x2BD7A6A6E99C2DDC)
+    # jump_const_2 = uint64(0x0992CCAF6A6FCA05)
+    # jump_const_3 = uint64(1)
+    # jump_init = uint64(0)
+    # long_2 = int64(2)
+    # long_64 = int64(64)
+
+    # @dpex.func
+    # def _xoroshiro128pp_jump(states, index):
+    #     """Advance the RNG in ``states[index]`` by 2**64 steps."""
+    #     s0 = jump_init
+    #     s1 = jump_init
+
+    #     for i in range(long_2):
+    #         if i == zero_idx:
+    #             jump_const = jump_const_1
+    #         else:
+    #             jump_const = jump_const_2
+    #         for b in range(long_64):
+    #             if jump_const & jump_const_3 << uint32(b):
+    #                 s0 ^= states[index, zero_idx]
+    #                 s1 ^= states[index, one_idx]
+    #             _xoroshiro128pp_next(states, index)
+
+    #     states[index, zero_idx] = s0
+    #     states[index, one_idx] = s1
+
+    jump_const_1 = uint64(0xDF900294D8F554A5)
+    jump_const_2 = uint64(0x170865DF4B3201FC)
     jump_const_3 = uint64(1)
     jump_init = uint64(0)
     long_2 = int64(2)
     long_64 = int64(64)
 
     @dpex.func
-    def _xoroshiro128p_jump(states, index):
+    def _xoroshiro128pp_jump(states, index):
         """Advance the RNG in ``states[index]`` by 2**64 steps."""
         s0 = jump_init
         s1 = jump_init
@@ -167,7 +215,7 @@ def create_xoroshiro128p_states(n_states, seed=None, subsequence_start=0, device
             else:
                 jump_const = jump_const_2
             for b in range(long_64):
-                if jump_const & (jump_const_3 << uint32(b)):
+                if jump_const & jump_const_3 << uint32(b):
                     s0 ^= states[index, zero_idx]
                     s1 ^= states[index, one_idx]
                 _xoroshiro128p_next(states, index)
@@ -189,13 +237,12 @@ def create_xoroshiro128p_states(n_states, seed=None, subsequence_start=0, device
                 "No CPU found, fallbacking random initiatlization to default " "device."
             )
 
-    states = dpctl.tensor.empty(
+    states = dpt.empty(
         sh=(n_states, 2),
         dtype=np.uint64,
         device=cpu_device if from_cpu_to_device else device,
     )
-
-    init_xoroshiro128p_states[1, 1](states)
+    init_xoroshiro128pp_states[1, 1](states)
 
     if from_cpu_to_device:
         return states.to_device(device)
@@ -204,9 +251,29 @@ def create_xoroshiro128p_states(n_states, seed=None, subsequence_start=0, device
         return states
 
 
-next_rot_1 = uint32(55)
-next_rot_2 = uint32(14)
-next_rot_3 = uint32(36)
+next_rot_pp_1 = uint32(17)
+next_rot_pp_2 = uint32(49)
+next_rot_pp_3 = uint32(21)
+next_rot_pp_4 = uint32(28)
+
+
+@dpex.func
+def _xoroshiro128pp_next(states, index):
+    """Return the next random uint64 and advance the RNG in states[index]."""
+    s0 = states[index, zero_idx]
+    s1 = states[index, one_idx]
+    result = _rotl(s0 + s1, next_rot_pp_1) + s0
+
+    s1 ^= s0
+    states[index, zero_idx] = _rotl(s0, next_rot_pp_2) ^ s1 ^ (s1 << next_rot_pp_3)
+    states[index, one_idx] = _rotl(s1, next_rot_pp_4)
+
+    return result
+
+
+next_rot_p_1 = uint32(24)
+next_rot_p_2 = uint32(16)
+next_rot_p_3 = uint32(37)
 
 
 @dpex.func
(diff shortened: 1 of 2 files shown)


fcharras:
@jjerphan to answer your suggestion in the other PR, I'd rather check that the RNG returns the exact same values than with a reliable reference implementation, than independently trying to test the good properties of randomness of the generator, which sounds harder.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+    return result
+
+
+@lru_cache
+def make_random_raw_kernel():
+    @dpex.kernel
+    def _get_random_raw_kernel(states, result):
+        result[zero_idx] = _xoroshiro128pp_next(states, zero_idx)
+
+    return _get_random_raw_kernel[1, 1]
```
jjerphan, line 41: Can you add docstrings to specify the kind of pseudo-random value this is generating (I assume float32 in [0, 1).)?

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+@lru_cache
+def make_rand_uniform_kernel_func(dtype):
```
jjerphan, line 45: Some comment regarding docstrings.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+    # `randomgen` that inspired this code contains errors.
+    # There's an issue here: https://github.com/bashtage/randomgen/issues/321
+    # This website https://prng.di.unimi.it/ seems to provide a ground truth
+    # implementation at https://prng.di.unimi.it/xoroshiro128plusplus.c but is not
+    # interfaced with python.
+    # We choose to mimic the reference implementation given in C even if there's no
+    # easy, reproducible set of instructions to test that it's working as expected.
+    # If the `randomgen` current implementation proves to be good, then the following
+    # block should be used instead, and the tests should be adapted. Else, the block
+    # can be removed.
```
jjerphan, line 205: We can easily come up with Python interface using Cython if needed. Let me know if this would be useful and I can work on this.

jjerphan (commented):
Some comments from a code and maintenance perspectives after a first walk in this bit fields' farm.

I need to get into the theoretical aspects first before continuing reviews.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+def make_rand_uniform_kernel_func(dtype):
+    if not hasattr(dtype, "name"):
+        raise ValueError(
+            "dtype is expected to have an attribute 'name', like np.dtype or numba "
+            "types."
+        )
+
+    if dtype.name == "float64":
+        convert_rshift = uint32(11)
+        convert_const = uint64(1) << uint32(53)
```
jjerphan, line 54: For consistency based on what follows.
```suggestion
        convert_const = float64(uint64(1) << uint32(53))
```

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+        convert_const_one = float64(1)
+
+        @dpex.func
+        def uint64_to_unit_float(x):
+            """Convert uint64 to float64 value in the range [0.0, 1.0)"""
+            return (x >> convert_rshift) * (convert_const_one / convert_const)
```
jjerphan, line 60: Ditto.
```suggestion
            return float64(x >> convert_rshift) * (convert_const_one / convert_const)
```

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+    #         else:
+    #             jump_const = jump_const_2
+    #         for b in range(long_64):
+    #             if jump_const & jump_const_3 << uint32(b):
+    #                 s0 ^= states[index, zero_idx]
+    #                 s1 ^= states[index, one_idx]
+    #             _xoroshiro128p_next(states, index)
+
+    #     states[index, zero_idx] = s0
+    #     states[index, one_idx] = s1
```
jjerphan, line 232: It looks like this part of the comment is duplicated and can be removed.

```suggestion
```

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+#     """Return the next random uint64 and advance the RNG in states[index]."""
+#     s0 = states[index, zero_idx]
+#     s1 = states[index, one_idx]
+#     result = s0 + s1
+
+#     s1 ^= s0
+#     states[index, zero_idx] = _rotl(s0, next_rot_p_1) ^ s1 ^ (s1 << next_rot_p_2)
+#     states[index, one_idx] = _rotl(s1, next_rot_p_3)
+
+#     return result
```
jjerphan, line 298: Is this apparently erroneous as well?

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+uint32_64 = uint32(64)
```
jjerphan, line 301: To me, this naming is clearer.
```suggestion
64_as_uint32 = uint32(64)
```

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+@dpex.func
+def _rotl(x, k):
+    """Left rotate x by k bits."""
+    return (x << k) | (x >> (uint32_64 - k))
```
jjerphan, line 307: If this rotation only works for numbers encoded on 32 bits, I think this must be documented.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,307 @@
...
+    return xoroshiro128pp_uniform
+
+
+def create_xoroshiro128pp_states(n_states, seed=None, subsequence_start=0, device=None):
+    """Returns a new device array initialized for n random number generators.
+
+    This initializes the RNG states so that each state in the array corresponds
+    subsequences in the separated by 2**64 steps from each other in the main
+    sequence.  Therefore, as long no thread requests more than 2**64 random numbers,
+    all of the RNG states produced by this function are guaranteed to be independent.
```
jjerphan, line 98:
```suggestion
    This initializes the RNG states so that states in the array corresponds
    to subsequences the separated by 2**64 steps from each other in the main
    sequence.
    Therefore, as long as no thread requests more than 2**64 random numbers,
    all the RNG states produced by this function are guaranteed to be independent.
```

jjerphan (commented):
Another pass for a few last comments.

sklearn_numba_dpex/common/random.py
fcharras: It works for 64 bits number, it's the shift amount that can be encoded as a `uint32` (it's an int ranging from 0 to 64)

IDK if there's a particular reason to force the `uint32` type for the shift amount but I'm refering to the same instructions in `numba.cuda.random`.

sklearn_numba_dpex/common/random.py
fcharras: Yes, this block is required by the erroneous block.

sklearn_numba_dpex/common/random.py
fcharras: There's not too much complexity in this code but that can give us more confidence in my copy pasting skills :grin:

sklearn_numba_dpex/common/random.py
fcharras: That's better for readability, but per casting rules we already have a `float64` and the comment already states that a `float64` is returned. This formula is used by all reference implementations so I'd also be afraid to add mistakes :-/

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,308 @@
...
+    #     for i in range(long_2):
+    #         if i == zero_idx:
+    #             jump_const = jump_const_1
+    #         else:
+    #             jump_const = jump_const_2
+    #         for b in range(long_64):
+    #             if jump_const & jump_const_3 << uint32(b):
+    #                 s0 ^= states[index, zero_idx]
+    #                 s1 ^= states[index, one_idx]
+    #             _xoroshiro128p_next(states, index)
```
fcharras, line 232: inline comment that hint at the difference there (xoroshirop vs xoroshiropp)

sklearn_numba_dpex/common/random.py
fcharras: add docstring that document x dtype

fcharras:
LGTM ?

sklearn_numba_dpex/common/tests/test_random.py
```diff
@@ -74,7 +82,65 @@ def test_xoroshiro128pp():
...
+        return
+
+    random_state = create_xoroshiro128pp_states(n_states=1, seed=42)
+    get_single_rand_value_kernel_64 = _make_get_single_rand_value_kernel(np.float64)
+    actual_res_64_to_32 = [
+        np.float32(get_single_rand_value_kernel_64(random_state))
+        for _ in expected_res_32
+    ]
+
+    np.testing.assert_allclose(actual_res_64_to_32, expected_res_32, rtol=1e-4)
```
jjerphan, line 146: Should we preferably use fixtures from scikit-learn for tests assertions (i.e. [`sklearn.utils._testing.assert*`](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_testing.py))?

jjerphan (approved):
LGTM. Just one comment regarding assertions.

sklearn_numba_dpex/common/tests/test_random.py
fcharras: I didn't want to add a dependency on sklearn but that's dumb sklearn is already a mandatory dependency :facepalm: will fix & merge

sklearn_numba_dpex/common/random.py
jjerphan: This was not entirely identical: `_xoroshiro128p_next` is called in this commented code, whereas `_xoroshiro128pp_next` is called, as indicated by @fcharras in a comment which was integrated in the meantime.

sklearn_numba_dpex/common/tests/test_random.py
jjerphan: Do not :facepalm: yourself: it's fine. :slightly_smiling_face:

ogrisel: I am reviewing this now. Please wait for my review before merging.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,309 @@
...
+
+    This initializes the RNG states so that states in the array correspond to
+    subsequences separated by 2**64 steps from each other in the main sequence.
+    Therefore, as long as no thread requests more than 2**64 random numbers, all the
+    RNG states produced by this function are guaranteed to be independent.
+
+    Parameters
+    ----------
+    n_states : int
+        number of RNG states to create
```
ogrisel, line 106: Could you please explain how to chose this value? The tests only use `n_states=1` if I am not mistaken.

I guess it has to do with the amount of parallelism you want to allow when generating a large number of random numbers?

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,309 @@
...
+
+    seed = uint64(seed)
+    n_states = int64(n_states)
+
+    init_const_1 = np.uint64(0)
+
+    @dpex.kernel
+    def init_xoroshiro128pp_states(states):
+        """
+        Use SplitMix64 to generate an xoroshiro128p state from 64-bit seed.
```
ogrisel, line 131:
```suggestion
        Use SplitMix64 to generate an xoroshiro128p state from a uint64 seed.
```

ogrisel (commented):
I find the tests in this PR quite unexpected, because they only ever generate one random number per kernel invocation which is really not the canonical usage for a GPU RNG implementation.

Could you please add a test that generate `1e7` values in parallel (with a typical value for `n_states`, not sure how to choose this depending on the device) and check that all the returned values are between `[0, 1)` and that binning in 10 bins using something like

```python
np.histogram(np.random.rand(int(1e7)), bins=np.linspace(0, 1, 11))[0]
```

returns around 1e6 (+/- 1e3) values per bin for a few arbitrary seeds and values for `n_states`?

This test should be parametrized both for `dtype=npfloat32` and `np.float64`.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,309 @@
...
+        @dpex.func
+        def uint64_to_unit_float(x):
+            """Convert uint64 to float32 value in the range [0.0, 1.0)
+            NB: this is different than original numba.cuda.random code. Instead of
+            generating a float64 random number before casting it to float32, a float32
+            number is generated from uint64 without intermediate float64. This enables
+            compatibility with devices that do not support float64 numbers.
+            However is seems to be exactly equivalent e.g it passes the float precision
+            test in sklearn.
+            """
```
ogrisel, line 78: PEP257 nitpick.

```suggestion
            """Convert uint64 to float32 value in the range [0.0, 1.0)

            NB: this is different than original numba.cuda.random code. Instead of
            generating a float64 random number before casting it to float32, a float32
            number is generated from uint64 without intermediate float64. This enables
            compatibility with devices that do not support float64 numbers.
            However is seems to be exactly equivalent e.g it passes the float precision
            test in sklearn.
            """
```

sklearn_numba_dpex/common/random.py
fcharras: > I guess it has to do with the amount of parallelism you want to allow when generating a large number of random numbers?

yes. n_states = n_threads

sklearn_numba_dpex/common/tests/test_random.py
```diff
@@ -0,0 +1,146 @@
...
+        1696280121849217124,
+        9039151786603216578,
+        1834868434787995627,
+        4294676528364711596,
+        5649342080498856832,
+    ]
+
+    actual_res_1 = [
+        int(dpt.asnumpy(get_random_raw(random_state))[0]) for _ in expected_res_1
+    ]
```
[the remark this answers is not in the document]
fcharras: > we should get the same 10 values, no?

we should, let's test this

fcharras:
> I find the tests in this PR quite unexpected

I wanted to avoid the compilcated test on rng quality and associated maintain burden, and only test that we have the same numbers than a reference implementation (here the implementation provided in `randomgen`). The first few numbers starting with the same seed should be enough to prove that the rng is rigorously identical to the reference ?

(there's a weakness in this plan with the issue on randomgen _jump_ function but at least it holds for the _next_ function)

sklearn_numba_dpex/common/random.py
ogrisel: Please make that explicit in the docstring for this parameter.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,309 @@
...
+# Reference resource about PRNG: https://prng.di.unimi.it/
+
+# NB2: original numba.cuda.random code also includes functions for generating normally
+# distributed floats but we don't include it here as long as it's not needed.
+
+zero_idx = int64(0)
+one_idx = int64(1)
+
+
+def get_random_raw(states):
```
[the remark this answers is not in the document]
ogrisel: Ok I did not know that. Let's keep this name and improve the docstring then.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,309 @@
...
+    """Returns one `uint64` random integer"""
```
ogrisel, line 30:
```suggestion
    """Returns one `uint64` random integer
    
    Similar to numpy.random.BitGenerator.random_raw but with parallelism
    controlled by the number of states.
    """
```

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,309 @@
...
+    result = dpt.empty(sh=(1,), dtype=np.uint64, device=states.device)
+    make_random_raw_kernel()(states, result)
+    return result
+
+
+@lru_cache
+def make_random_raw_kernel():
```
ogrisel, line 37:
```suggestion
def make_random_raw_kernel():
    """Returns random `uint64` integers
    
    Similar to numpy.random.BitGenerator.random_raw but with parallelism
    controlled by the number of states.
    """
```

sklearn_numba_dpex/common/random.py
fcharras: The number of threads [is 2**64 ](https://prng.di.unimi.it/)so we should be fine :grin: 

image

fcharras:
I pushed the suggestions / answered the comments, will work on the finer test for parallel rng tomorrow.

fcharras:
doesn't 1e7 sound a lot for a unitt test ran in a github action runner ? maybe there's a smaller setup that is still enough ?

sklearn_numba_dpex/common/random.py
ogrisel: > and really wouldn't work with parallelism. It's not fit for parallel calls...

Yes I realized that after posting my comment and updated the suggestion to only refer to the numpy name.

ogrisel:
> doesn't 1e7 sound a lot for a unitt test ran in a github action runner ? maybe there's a smaller setup that is still enough ?

As you wish. Maybe 1e6 is enough and you can reduce the number of bins, e.g. to `np.linspace(0, 1, 6)` to get 2e5 +/- 1e3 in each bin or something.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,312 @@
...
+            else:
+                jump_const = jump_const_2
+            for b in range(long_64):
+                if jump_const & jump_const_3 << uint32(b):
+                    s0 ^= states[index, zero_idx]
+                    s1 ^= states[index, one_idx]
+                _xoroshiro128pp_next(states, index)
+
+        states[index, zero_idx] = s0
+        states[index, one_idx] = s1
```
ogrisel, line 199:
```suggestion
    def _xoroshiro128pp_jump(states, state_idx):
        """Advance the RNG in ``states[state_idx]`` by 2**64 steps."""
        s0 = jump_init
        s1 = jump_init

        for i in range(long_2):
            if i == zero_idx:
                jump_const = jump_const_1
            else:
                jump_const = jump_const_2
            for b in range(long_64):
                if jump_const & jump_const_3 << uint32(b):
                    s0 ^= states[state_idx, zero_idx]
                    s1 ^= states[state_idx, one_idx]
                _xoroshiro128pp_next(states, state_idx)

        states[state_idx, zero_idx] = s0
        states[state_idx, one_idx] = s1
```

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,312 @@
...
+    """Return the next random uint64 and advance the RNG in states[index]."""
+    s0 = states[index, zero_idx]
+    s1 = states[index, one_idx]
+    result = _rotl(s0 + s1, next_rot_1) + s0
+
+    s1 ^= s0
+    states[index, zero_idx] = _rotl(s0, next_rot_2) ^ s1 ^ (s1 << shift_1)
+    states[index, one_idx] = _rotl(s1, next_rot_3)
+
+    return result
```
ogrisel, line 303:
```suggestion
@dpex.func
def _xoroshiro128pp_next(states, state_idx):
    """Return the next random uint64 and advance the RNG in states[index]."""
    s0 = states[state_idx, zero_idx]
    s1 = states[state_idx, one_idx]
    result = _rotl(s0 + s1, next_rot_1) + s0

    s1 ^= s0
    states[state_idx, zero_idx] = _rotl(s0, next_rot_2) ^ s1 ^ (s1 << shift_1)
    states[state_idx, one_idx] = _rotl(s1, next_rot_3)

    return result
```

ogrisel (commented):
More feedback based on my updated understanding.

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -0,0 +1,106 @@
...
+    argmin_reduction_1d_kernel = make_argmin_reduction_1d_kernel(
+        size=len(array_in),
+        work_group_size=work_group_size,
+        device=device,
+        dtype=dtype,
+    )
+
+    actual_result = dpt.asnumpy(argmin_reduction_1d_kernel(array_in))[0]
+    assert actual_result.dtype == np.int32
+    assert actual_result == expected_result
```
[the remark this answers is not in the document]
fcharras: github is bugged because of the conflict after the merge of the reduction kernels, the diff will display properly when main is pulled. (I've opened a lot of cascades of PRs that rely on each other to make review and step-by-step merge easier)

sklearn_numba_dpex/common/random.py
fcharras: About the scope of this PR the purpose here is not to provide an high-level interface dedicated to user that just generate random number like e.g `numpy.random.rand` would do. For me it was only about serving the need of the  kmeans++ where the device functions (those defined with `@dpex.func`) are used in to have rng in each thread fused with other instructions. It is a lower level use where it is up to the user to choose carefully `n_states` and then use it accordingly in other kernels. And we're not preventing bad use, like having several threads using the same state. Maybe I should have made it clearer in the description of the PR so you can start with a better understanding a priori.

(also, there is rng with better user interface, e.g in jax each thread use a different hash function on integers)

That being said I think your docstring is fine and we can even omit to document the constraint of not having more than `2**64` threads iterate on more than `2**64` values since those are virtually impossible to reach ?

> You mean the maximum number of threads that we can use safely?
yes, the algorithm is safe as long as `n_states < 2**64` and each state not updated more than `2**64` times.

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
fcharras: Actually not bugged, the cascade of PR was set such that reduction kernels was merged on this branch. Fixed that and merged them to master.

fcharras:
Had some troubles writing the rng quality test but here it is + applied all other suggestions. TY for the docstrings :smile:

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,312 @@
...
+    Parameters
+    ----------
+    n_states : int
+        number of RNG states to create
+
+    seed : int or None
+        starting seed for the list of generators
+
+    subsequence_start : int
+        advance the first RNG state by a multiple of 2**64 steps
```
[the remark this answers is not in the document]
fcharras: Actually restoring it, a test rightly uses it (to test the jump function).

sklearn_numba_dpex/common/tests/test_random.py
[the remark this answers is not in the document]
ogrisel: This comment has not yet been addressed I think.

sklearn_numba_dpex/common/random.py
ogrisel: This suggestion was not considered yet.

sklearn_numba_dpex/common/random.py
```diff
@@ -0,0 +1,340 @@
...
+@lru_cache
+def make_rand_uniform_kernel_func(dtype):
+    """Instantiate a kernel function that returns a random float in [0, 1)
+
+    This factory returns a kernel function specialized to either generate
+    float32 or float64 values. Care has been taken so that the float32
+    variant can be compiled to target a device that does not support the
+    float64 aspect.
+
+    The returned kernel functions takes two arguments:
```
ogrisel, line 59:
```suggestion
    The returned kernel function takes two arguments:
```

ogrisel (approved):
There are a few suggestions left in the diff view of the PR:

https://github.com/soda-inria/sklearn-numba-dpex/pull/45/files

pending those. LGTM!

sklearn_numba_dpex/common/tests/test_random.py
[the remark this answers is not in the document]
fcharras: I added it in the already existing unittest

PR #46, opened 2022-10-27
Extend sum reduction kernel, add argmin reduction kernel

fcharras:
Two things in this PR:

- [x] extend the sum reduction kernel for 1d arrays to 2d arrays when reducing over axis 1
- [x] add a similar argmin kernel

Not sure what is better, using those kernels or using `dpnp` functions.

Pros for using those kernels:
- limit the dependency to `dpnp` (which can especially be a good things if `numba_dpex` get interoperability to nvidia/amd gpus before `dpnp` does)
- more confidence regarding the strategy for dispatching to thread that fit more GPU devices, it's not clear at what point `dpnp` can run well on gpus
- fun

Cons for using those kernels:
- it's verbose, and if we want to optimize other usecases (depending if axis0 is "big enough", or else which is bigger from axis 0 or axis 1,...) it would require yet other kernels.
- it does not sound like it's a good thing to jit those simple operations, the added jit time might get annoying
- at some point, it can be assumed that a reliable tensor library (probably `dpnp` ?) will be available and the kernels will be obsolete anyway.
- even if the implementation fits the gpu model, I think the performance is not that great, it might become competitive if there's a public api in `numba_dpex` to use async dispatching like described in https://github.com/IntelPython/numba-dpex/issues/769 .
- in fact, the performance does not even really matter because until now I don't think it contributes much to total execution time anyway.

Those pro / cons can also be weighted for all other kernels in this file, implementations for most of those are already available in dpnp.

### DIFF
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -131,9 +131,11 @@ def make_half_l2_norm_2d_axis0_kernel(size0, size1, work_group_size, dtype):
 
 @lru_cache
 def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dtype):
-    """numba_dpex does not provide tools such as `cuda.reduce` so we implement from
-    scratch a reduction strategy. The strategy relies on the commutativity of the
-    operation used for the reduction, thus allowing to reduce the input in any order.
+    """Implement data_2d.sum(axis=1) or data_1d.sum()
+
+    numba_dpex does not provide tools such as `cuda.reduce` so we implement from scratch
+    a reduction strategy. The strategy relies on the commutativity of the operation used
+    for the reduction, thus allowing to reduce the input in any order.
 
     The strategy consists in performing local reductions in each work group using local
     memory where each work item combine two values, thus halving the number of values,
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -24,6 +24,8 @@ import numpy as np
 import dpctl.tensor as dpt
 import numba_dpex as dpex
 
+from sklearn_numba_dpex.common._utils import check_power_of_2
+
 
 zero_idx = np.int64(0)
 
@@ -160,6 +162,8 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
     the same order of magnitude than `preferred_work_group_size_multiple`. If not,
     other reduction implementations might give better performances.
     """
+    check_power_of_2(work_group_size)
+
     # Number of iteration in each execution of the kernel:
     local_n_iterations = np.int64(math.floor(math.log2(work_group_size)) - 1)
 
@@ -313,6 +317,11 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
         global_size = n_groups * work_group_size
         kernel = partial_sum_reduction[global_size, work_group_size]
         result_shape = n_groups if is_1d else (n_rows, n_groups)
+        # NB: here memory for partial results is allocated ahead of time and will only
+        # be garbage collected when the instance of `sum_reduction` is garbage
+        # collected. Thus it can be more efficient to re-use a same instance of
+        # `sum_reduction` (e.g within iterations of a loop) since it avoid deallocation
+        # and reallocation every time.
         result = dpt.empty(result_shape, dtype=dtype, device=device)
         kernels_and_empty_tensors_pairs.append((kernel, result))
 
@@ -329,6 +338,8 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
 @lru_cache
 def make_argmin_reduction_1d_kernel(size, work_group_size, device, dtype):
     """Implement 1d argmin with the same strategy than for make_sum_reduction_2d_axis1_kernel."""
+    check_power_of_2(work_group_size)
+
     # Number of iteration in each execution of the kernel:
     local_n_iterations = np.int64(math.floor(math.log2(work_group_size)) - 1)
 
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -147,30 +147,32 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
     `2 * work_group_size`. This is repeated as many time as needed until only one value
     remains in global memory.
 
-    NB: work_group_size is assumed to be a power of 2.
+    Notes
+    -----
+    `work_group_size` is assumed to be a power of 2.
 
-    NB: if size1 is None then the kernel expects 1d tensor inputs. If size1 is not None
-    then the expected shape of input tensors is (size0, size1), and the reduction
+    if `size1` is None then the kernel expects 1d tensor inputs. If `size1` is not None
+    then the expected shape of input tensors is `(size0, size1)`, and the reduction
     operation is equivalent to input.sum(axis=1). In this case, the kernel is a good
-    choice if size1 >> preferred_work_group_size_multiple, and if size0 ranges in the
-    same order of magnitude than preferred_work_group_size_multiple. If not, other
-    reduction implementations might give better performances.
-
-    ???: how does this strategy compares to having each thread reducing N contiguous
-    items ?
+    choice if `size1` >> `preferred_work_group_size_multiple`, and if `size0` ranges in
+    the same order of magnitude than `preferred_work_group_size_multiple`. If not,
+    other reduction implementations might give better performances.
     """
     # Number of iteration in each execution of the kernel:
     local_n_iterations = np.int64(math.floor(math.log2(work_group_size)) - 1)
 
     zero = dtype(0.0)
-    two_long = np.int64(2)
     one_idx = np.int64(1)
+    minus_one_idx = np.int64(-1)
+    two_as_a_long = np.int64(2)
 
     is_1d = size1 is None
+    # TODO: this set of kernel functions could be abstracted away to other coalescing
+    # functions
     if is_1d:
         sum_axis_size = size0
         n_rows = np.int64(1)
-        local_data_size = work_group_size
+        local_values_size = work_group_size
 
         @dpex.func
         def set_col_to_zero(array, i):
@@ -181,7 +183,7 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
             to_array[to_col] = from_array[from_col]
 
         @dpex.func
-        def coalesce_cols(
+        def add_cols(
             from_array,
             left_from_col,
             right_from_col,
@@ -191,7 +193,7 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
             to_array[to_col] = from_array[left_from_col] + from_array[right_from_col]
 
         @dpex.func
-        def coalesce_cols_inplace(
+        def add_cols_inplace(
             array,
             from_col,
             to_col,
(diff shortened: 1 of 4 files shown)


sklearn_numba_dpex/common/kernels.py
```diff
@@ -148,6 +148,16 @@ def make_sum_reduction_1d_kernel(size, work_group_size, device, dtype):
     remains in global memory.
 
     NB: work_group_size is assumed to be a power of 2.
+
+    NB: if size1 is None then the kernel expects 1d tensor inputs. If size1 is not None
+    then the expected shape of input tensors is (size0, size1), and the reduction
+    operation is equivalent to input.sum(axis=1). In this case, the kernel is a good
+    choice if size1 >> preferred_work_group_size_multiple, and if size0 ranges in the
+    same order of magnitude than preferred_work_group_size_multiple. If not, other
+    reduction implementations might give better performances.
```
jjerphan, line 157: I guess you can add a ["Notes" section following numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#notes)?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -156,21 +166,107 @@ def make_sum_reduction_1d_kernel(size, work_group_size, device, dtype):
...
+        @dpex.func
+        def set_col_to_zero(array, i):
+            array[i] = zero
+
+        @dpex.func
+        def copy_col(from_array, from_col, to_array, to_col):
+            to_array[to_col] = from_array[from_col]
+
+        @dpex.func
+        def coalesce_cols(
```
jjerphan, line 184: Since the coalescences are additions only, would it make sense to use `add_*_cols_*` instead of `coalesce_*_cols_*`?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -156,21 +166,107 @@ def make_sum_reduction_1d_kernel(size, work_group_size, device, dtype):
...
+
+        @dpex.func
+        def coalesce_first_cols(from_array, to_array, to_col):
+            for row in range(n_rows):
+                to_array[row, to_col] = (
+                    from_array[row, zero_idx] + from_array[row, one_idx]
+                )
+
+    two_long = np.int64(2)
+    m_one_idx = np.int64(-1)
```
jjerphan, line 250: Shall we be verbose?
```suggestion
    two_as_a_long = np.int64(2)
    minus_one_idx = np.int64(-1)
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -215,13 +309,122 @@ def partial_sum_reduction(
...
+                local_argmin[local_work_id] = x_idx
+                local_data[local_work_id] = data[x_idx]
+
+            else:
+                if has_previous_result:
+                    y_idx = previous_result[y_idx]
+
+                x_data = data[x_idx]
+                y_data = data[y_idx]
+                if x_data <= y_data:
```
jjerphan, line 382: This is closer to the tie-break strategies in scikit-learn.
```suggestion
                if x_data < y_data or (x_data == y_data and x_idx < y_idx):
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -215,13 +309,122 @@ def partial_sum_reduction(
...
+    # TODO: the first call of partial_argmin_reduction in the final loop should be
+    # written with only two arguments since "previous_result" does not exist yet.
+    # It seems it's not possible to get a good factoring of the code to avoid copying
+    # most of the code for this with @dpex.kernel, for now we resort to branching.
+    @dpex.kernel
+    # fmt: off
+    def partial_argmin_reduction(
+        data,               # IN        (size,)
+        previous_result,    # IN        (current_size,)
+        result,             # OUT       (math.ceil(
```
jjerphan, line 345: Can we give a semantic to `results`? For instance:
```suggestion
        argmin_indices,     # OUT       (math.ceil(
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -215,13 +309,122 @@ def partial_sum_reduction(
...
+    inf = dtype(np.inf)
+
+    # TODO: the first call of partial_argmin_reduction in the final loop should be
+    # written with only two arguments since "previous_result" does not exist yet.
+    # It seems it's not possible to get a good factoring of the code to avoid copying
+    # most of the code for this with @dpex.kernel, for now we resort to branching.
+    @dpex.kernel
+    # fmt: off
+    def partial_argmin_reduction(
+        data,               # IN        (size,)
```
jjerphan, line 343: Is this a bit more precise?
```suggestion
        values,               # IN        (size,)
```

jjerphan (commented):
Thanks @fcharras. Here are a few comments.

sklearn_numba_dpex/common/kernels.py
fcharras: I liked `coalesce` here because it stresses the fact that the kernel could be abstracted to and extended to other functions that can be accumulated (like min, max, etc), just by replacing the "sum" op

jjerphan: I see.

I think we started using `reduce` and `reduction` as terms commonly. Is this semantic different?

For now, I am +0 regarding creating private abstractions for coalescences (or reductions) at this level mainly because YAGNI and because private renaming are cheap.

fcharras: I only intended to hint into the possibility of this abstraction, not implement it now. But I agree with your point I'll just revert to naming to `add_*_cols_*` and explicit the hint in a comment.

fcharras:
TODO: add couple of tests and merge

jjerphan (approved):
As discussed on a call with @fcharras, this is ready to merge modulo a few tests on the new common kernels.

fcharras:
Tests added. LGTM ?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -215,13 +310,122 @@ def partial_sum_reduction(
...
-        result = dpt.empty(n_groups, dtype=dtype, device=device)
+        result_shape = n_groups if is_1d else (n_rows, n_groups)
+        result = dpt.empty(result_shape, dtype=dtype, device=device)
         kernels_and_empty_tensors_pairs.append((kernel, result))
 
     def sum_reduction(summands):
+        # TODO: manually dispatch the kernels with a SyclQueue
         for kernel, result in kernels_and_empty_tensors_pairs:
             kernel(summands, result)
             summands = result
```
ogrisel, line 321: It would be more efficient to do the calls to `dpt.empty(result_shape, dtype=dtype, device=device)` lazily inside this loop, instead of allocating all the results buffers ahead of time?

Or maybe the Python GC would add sequential overhead that would kill the performance?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -147,66 +147,161 @@ def make_sum_reduction_1d_kernel(size, work_group_size, device, dtype):
     `2 * work_group_size`. This is repeated as many time as needed until only one value
     remains in global memory.
 
-    NB: work_group_size is assumed to be a power of 2.
+    Notes
+    -----
+    `work_group_size` is assumed to be a power of 2.
```
ogrisel, line 152: What happens if it's not the case? Do you get an error or a invalid result?

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -0,0 +1,92 @@
...
+)
+from sklearn_numba_dpex.testing.config import float_dtype_params
+
+
+@pytest.mark.parametrize("dtype", float_dtype_params)
+def test_sum_reduction_2d(dtype):
+    n_reductions = 3
+    n_items = 4
+    device = dpctl.SyclDevice()
+    work_group_size = device.max_work_group_size
```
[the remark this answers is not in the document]
ogrisel: You can use random data values from numpy that you move to the sycl device via `dpt` instead of using `dpt.reshape(dpt.arange(12, dtype=dtype), (3, 4))` and then compare to the reduction results computed by numpy.

sklearn_numba_dpex/common/kernels.py
fcharras: By allocating ahead of time we also keep the buffers allocated and re-use the same buffers across all future calls to `sum_reduction`. The buffers will be only be garbage collected when `sum_reduction` is garbage collected. A given instance of `sum_reduction` in our loops is going to be called once per iteration so I think it's more sensible this way.

sklearn_numba_dpex/common/kernels.py
fcharras: an invalid result

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -0,0 +1,92 @@
...
+    expected_result = 6
+
+    actual_result = int(dpt.asnumpy(sum_reduction_1d_kernel(array_in))[0])
+
+    assert actual_result == expected_result
+
+
+@pytest.mark.parametrize("dtype", float_dtype_params)
+def test_argmin_reduction_1d(dtype):
+    n_items = 4
```
jjerphan, line 68: Would it make sense to define `n_items` based on the length of `array_in`?

This comment also applies in other tests.

jjerphan (approved):
LGTM, @ogrisel outlined all everything already.

I guess, `make_sum_reduction_2d_axis0_kernel` can be implemented when need and those test generalised if it is the case. What do you think?

jjerphan:
I meant that `make_sum_reduction_2d_axis1_kernel` is implemented as it is needed for `kmeans++`, but `make_sum_reduction_2d_axis0_kernel` is not at the moment but might be needed in the future.

If implemented in the future, tests could be adapted accordingly; nothing is to change for this PR.

ogrisel:
Let's not worry about `make_sum_reduction_2d_axis0_kernel` before we ever need it and refactor only then if needed.

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
fcharras: Why use random data values for a sum/argmin ? I can't think of specific cases that we could miss with hard-coded trivial data, and it's not readable. Plus those tests can be a little expensive because there's a jit for each (n_reduction, n_items) combination.

OK for tests that highlight what happens with invalid combinations but I don't think it's very important either since those functions are not intended to be public but only have use in top level functions that are also tested.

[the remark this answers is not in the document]
fcharras: Also when work_group_size is not a power of 2 the behavior should be undefined. Could work, could fail. could segfault. Do you want that the instantiation function perform some input validation and raise error if `work_group_size` is not a power of 2 ?

sklearn_numba_dpex/common/kernels.py
ogrisel: I think we should try to detect it then :)

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
ogrisel: I think it's should be quite cheap to check yes + it's protected by the lru cache.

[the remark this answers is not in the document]
ogrisel: > OK for tests that highlight what happens with invalid combinations but I don't think it's very important either since those functions are not intended to be public but only have use in top level functions that are also tested.

Regular data or deterministically RNG'ed data as you prefer. My point it to do some checks with larger dimensions than `workgroup_size` and compare to the results of the same reduction computed by numpy.

sklearn_numba_dpex/common/kernels.py
ogrisel: Alright!

ogrisel: Maybe add an inline comment to explain this!

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
fcharras: OK for validating `work_group_size`, it's actually very cheap.

I'll parametrize work groups and and a use case with shape of axis 1 set to length 5 to test cases where work group size and axis sizes do not divide each other

ogrisel (approved):
LGTM!

fcharras:
Merged! ty for review

PR #48, opened 2022-11-04
Refacto: get data conversion steps more visible

fcharras:
The goal now that our KMeans is complete is to work toward enabling the use of the estimators with the `sklearn_numba_dpex` engine with inputs that are not numpy arrays but dpnp arrays or dpt tensors, with a satisfying level of input validation, and while keeping the level of features that we currently have (in particular testing against sklearn unit tests).

It is harder than it sounds and I've been running in circle trying to achieve this, I'll work toward a sequence of small PRs to make data conversions easier to track, by:

- decoupling data validation/conversion steps and actual compute
- leveraging "compute follow data" model to minimize the number of occurences of data parameters in the pipeline

This PR is a small refactoring of the KMeans driver where the data conversion steps are extracted from the driver functions.

### DIFF
--- a/sklearn_numba_dpex/kmeans/drivers.py
+++ b/sklearn_numba_dpex/kmeans/drivers.py
@@ -169,13 +169,7 @@ class KMeansDriver:
         self.has_aspect_fp64 = device_params.has_aspect_fp64
 
     def lloyd(
-        self,
-        X,
-        sample_weight,
-        centers_init,
-        max_iter=300,
-        verbose=False,
-        tol=1e-4,
+        self, X, sample_weight, centers_init, max_iter=300, verbose=False, tol=1e-4
     ):
         """This call is expected to accept the same inputs than sklearn's private
         _kmeans_single_lloyd and produce the same outputs.
@@ -187,9 +181,7 @@ class KMeansDriver:
         use_uniform_weights = (sample_weight == sample_weight[0]).all()
 
         X_t, sample_weight, centroids_t = self._load_transposed_data_to_device(
-            X,
-            sample_weight,
-            cluster_centers,
+            X, sample_weight, cluster_centers
         )
 
         (assignments_idx, inertia, best_centroids, n_iteration,) = self._lloyd(
@@ -262,10 +254,7 @@ class KMeansDriver:
         )
 
         compute_inertia_kernel = make_compute_inertia_kernel(
-            n_samples,
-            n_features,
-            self.work_group_size,
-            compute_dtype,
+            n_samples, n_features, self.work_group_size, compute_dtype
         )
 
         reset_cluster_sizes_private_copies_kernel = make_initialize_to_zeros_2d_kernel(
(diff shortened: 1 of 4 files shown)


sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -734,28 +726,30 @@ def _get_labels_inertia(
...
 
-    def _get_euclidean_distances(self, X, Y):
-        (X, _, Y, compute_dtype, output_dtype, work_group_size,) = self._check_inputs(
-            X,
-            sample_weight=_IgnoreSampleWeight,
-            cluster_centers=Y,
+        euclidean_distances = self._get_euclidean_distances(
+            X_t,
+            Y_t,
         )
```
jjerphan, line 745:
```suggestion
        euclidean_distances = self._get_euclidean_distances(X_t, Y_t)
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -799,11 +790,6 @@ def _set_dtype(self, X, sample_weight, centers_init):
                 f"will default back to float32 type."
             )
             compute_dtype = np.float32
-        elif compute_dtype != X.dtype:
-            text = (
-                f"KMeans is set to run with dtype {compute_dtype} but the data has "
-                f"been submitted with type {X.dtype}."
-            )
```
jjerphan, line 806: I think we need to add a test that now access that `compute_dtype != X.dtype` is supported.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -166,19 +168,6 @@ def __init__(
...
-            if (dtype != np.float32) and (dtype != np.float64):
-                raise ValueError(f"Valid types are float64, float32, but got f{dtype}")
-            self.dtype = dtype
-            if (self.dtype == np.float64) and not self.has_aspect_fp64:
-                raise NotSupportedByEngineError(
-                    f"Computations with precision f{self.dtype} has been explicitly "
-                    f"requested to the KMeans driver but the device {dpctl_device.name} does not "
-                    f"support it."
-                )
-
```
jjerphan, line 181: Should part of this previous validation move to the current `_check_inputs`?

jjerphan (commented):
Thanks for this decoupling, @fcharras.

Here are a few comments, questions and suggestions.

sklearn_numba_dpex/kmeans/drivers.py
fcharras: Indeed I think inlining is nicer in general ? I didn't realize that it's the final comma that triggers `black` indentation. There are a lot of other occurrences of indented lists of variables that can be inlined.

sklearn_numba_dpex/kmeans/drivers.py
fcharras: Answering this comment and the following: the driver exposes a "dtype" option to force the `compute_dtype` no matter what the dtype of the input is. In practice it is used nowhere. We used in the benchmark in the very early commits of the repo.

What I propose here is to remove this option entirely and only set the driver to use for compute the `dtype` the data has been submitted. (except only for the case where the dtype is not supoported by the device)

sklearn_numba_dpex/kmeans/drivers.py
fcharras: I add inlining for almost all occurences, let me know what you think.

jjerphan (approved):

PR #49, opened 2022-11-09
ENH Performance fixes and improvements

fcharras:
Fixes https://github.com/soda-inria/sklearn-numba-dpex/issues/52

Several issues are addressed in this PR that could cause loss of performance and also cause misinterpretation of benchmarks (and result in a 10-20% performance increase overall).

### Use a more sensible size for the window over centroids

Following https://github.com/IntelPython/dpctl/pull/894 I see that using `preferred_work_group_size_multiple` from `pyopencl` as an heuristic to set the size of the window was misled. The PR replaces with another heuristic that estimates the "sub_group_size" (a.k.a warp size) (see https://github.com/IntelPython/dpctl/issues/975 ).

As pointed out by the original paper the width of the window is better set equal to the warp size. It's good that it's a multiple of the warp size to optimize memory bandwith, and at the same time it's good to keep it small to avoid overusing the private memory (since each thread allocates a number of elements in private memory equal to the width of the window).

The height of the window is chosen to that each item in a work group is responsible for loading one and only one element into the window, which should also optimize memory bandwith. 

### Set work group size to max_work_group_size

The two previous points ensure that increasing work group size does not cause loss of perfornance:

- does not cause extra compute on padding
- does not cause registry spilling since the width of the window is fixed to the sub group size in any case

so that the work group size can be set to max_work_group_size, which is generally a good thing as long as there isn't concerns about registry spilling, saturating barrier registries, or local memory (and this PR extends it to all other side kernels).

A nice side effect after this PR is that the heuristic should apply equally to all devices and proposes to remove other parameters affecting width and height of the window (which on a second look doesn't look useful if the only effect is balancing the number of iterations from inner loops to outer loops).


fcharras:
(ping @oleksandr-pavlyk , you might be interested by those changes if you're also trying to reproduce this code in C++)

fcharras:
I moved the fix part of this PR to another PR https://github.com/soda-inria/sklearn-numba-dpex/pull/51/files

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
```diff
@@ -15,16 +17,18 @@ def make_compute_euclidean_distances_fixed_window_kernel(
...
     n_features,
     n_clusters,
-    preferred_work_group_size_multiple,
-    centroids_window_width_multiplier,
-    centroids_window_height,
+    sub_group_size,
     work_group_size,
     dtype,
 ):
+    centroids_window_height = work_group_size // sub_group_size
```
ogrisel, line 24: Could you please add an inline comment to explain this?

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -301,7 +297,7 @@ def fused_lloyd_single_step(
         weight = sample_weight[sample_idx]
 
         privatization_idx = (
-            sample_idx // preferred_work_group_size_multiple
+            sample_idx // sub_group_size
         ) % n_centroids_private_copies
```
ogrisel, line 301: Could you please insert an inline comment to explain this?

Also doesn't black makes this fit on one line? Maybe not.

ogrisel (approved):
I like the code simplification even though I do not understand the details :)

But I ran the benchmark on my local laptop with an oldish mobile Intel GPU and the new code is as fast as the best case of the previous code `work_group_size_multiplier=4` or work_group_size_multiplier=2` in my case.

So +1 for me pending the following nitpicks:

fcharras:
Ty for review. NB: for the moment this PR points at the branch that is also an open PR https://github.com/soda-inria/sklearn-numba-dpex/pull/51

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
```diff
@@ -4,7 +4,9 @@
 import numpy as np
 import numba_dpex as dpex
 
-from ._base_kmeans_kernel_funcs import make_pairwise_ops_base_kernel_funcs
+from ._base_kmeans_kernel_funcs import (
+    make_pairwise_ops_base_kernel_funcs,
+)
```
jjerphan, line 9: Nit: this better be reverted if nothing has changed not to obfuscate the history.

```suggestion
from ._base_kmeans_kernel_funcs import make_pairwise_ops_base_kernel_funcs
```

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -203,39 +203,34 @@ def fused_lloyd_single_step(
...
 
                 accumulate_dot_products(
-                    sample_idx,
-                    first_feature_idx,
-                    X_t,
-                    centroids_window,
-                    dot_products,
-                    is_last_feature_window,
-                    is_last_centroid_window
+                    sample_idx, first_feature_idx, X_t, centroids_window, dot_products, is_last_feature_window, is_last_centroid_window
```
jjerphan, line 224: Can this change be reverted?

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
```diff
@@ -15,16 +17,18 @@ def make_compute_euclidean_distances_fixed_window_kernel(
...
+    sub_group_size,
     work_group_size,
     dtype,
 ):
+    centroids_window_height = work_group_size // sub_group_size
+    if centroids_window_height * sub_group_size != work_group_size:
+        raise ValueError(
+            "Expected work_group_size to be a multiple of sub_group_size but got "
+            f"sub_group_size={sub_group_size} and work_group_size={work_group_size}"
+        )
```
jjerphan, line 29: Can you please add a test to assert this error is raised when it is appropriate?

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -18,16 +19,18 @@ def make_label_assignment_fixed_window_kernel(
...
+    sub_group_size,
     work_group_size,
     dtype,
 ):
+    centroids_window_height = work_group_size // sub_group_size
+    if centroids_window_height * sub_group_size != work_group_size:
+        raise ValueError(
+            "Expected work_group_size to be a multiple of sub_group_size but got "
+            f"sub_group_size={sub_group_size} and work_group_size={work_group_size}"
+        )
```
jjerphan, line 31: Same comment: can you please add a test to assert this error is raised when it is appropriate?

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -104,30 +105,26 @@ def assignment(
...
 
                 dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
 
                 accumulate_dot_products(
-                    sample_idx,
-                    first_feature_idx,
-                    X_t,
-                    centroids_window,
-                    dot_products, is_last_feature_window, is_last_centroid_window
+                    sample_idx, first_feature_idx, X_t, centroids_window, dot_products, is_last_feature_window, is_last_centroid_window
```
jjerphan, line 121: Can this be `blackified`?

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
```diff
@@ -82,32 +84,35 @@ def compute_distances(
...
-                    is_last_feature_window
                 )
 
                 dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
-                accumulate_sq_distances(sample_idx, first_feature_idx, X_t, centroids_window, sq_distances, is_last_feature_window, is_last_centroid_window)
 
-                dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+                accumulate_sq_distances(
+                    sample_idx, first_feature_idx, X_t, centroids_window, sq_distances, is_last_feature_window, is_last_centroid_window
+                )
```
jjerphan, line 101: Can this be `blackified`?

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
```diff
@@ -82,32 +84,35 @@ def compute_distances(
...
 
                 first_feature_idx += centroids_window_height
 
+                dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
```
jjerphan, line 105: Are there any particular motivation for moving barrier in kernels, such as this one?

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -203,39 +203,34 @@ def fused_lloyd_single_step(
...
                 is_last_feature_window = _1 == last_feature_window_idx
+
+                # centroids_window is modified inplace
                 load_window_of_centroids_and_features(
                     first_feature_idx,
                     loading_centroid_idx,
                     window_loading_centroid_idx,
                     window_loading_feature_offset,
                     current_centroids_t,
                     centroids_window,
```
jjerphan, line 215: Do you might this PR might be the opportunity to introduce the hint we once talked about during a call?
```suggestion
                load_window_of_centroids_and_features(
                    first_feature_idx,
                    loading_centroid_idx,
                    window_loading_centroid_idx,
                    window_loading_feature_offset,
                    current_centroids_t,
                    # OUT
                    centroids_window,
```

jjerphan (commented):
Thanks, @fcharras.

A few comments, and this LGTM!

fcharras:
Ty for review! but part of the diff in this PR is specific to the branch that is in PR here https://github.com/soda-inria/sklearn-numba-dpex/pull/51

Hence It would make more sense to review https://github.com/soda-inria/sklearn-numba-dpex/pull/51 first

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
fcharras: This is blackified already. black does not always try to fit the max line length requirement. If you want to hard enforce this another PR should be opened to configure it properly imo.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
fcharras: I autoformat all my files. doesn't our test pipeline test for conformity to black anyway ?

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
fcharras: I find it more sensible to set the barrier to the latest possible instruction (it maybe gives more room for the threads to synchronize) but in fact I don't think it really matters

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
fcharras: this change in particular is actually nicer do you mind if I automerge it directly on main

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
fcharras: I can maybe open a separate PR to add the hint in all kernel calls of the driver file

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
fcharras: it seems that in some block in some files black just refuses to apply :thinking:

fcharras: about the inline comment, it's already here, i reformated a bit and updated to replace preferred_work_group_size with sub_group_size

jjerphan:
I am waiting for #51 to be merged in `main`, and then `main` to be merged in the branch of this PR before re-reviewing it.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
jjerphan: Yes, probably doing it in another PR is more convenient. :+1:

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
fcharras: It's added in `lloyd_single_step.py`

fcharras:
:sweat_drops: there was a bit of a mess with the entangled pr that were merged recently, everything is back to order I think. 

I think comments have been adressed, and the same changes have been applied to kmeans++.

pyopencl and the device module are not needed anymore, and removed.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -202,27 +219,30 @@ def fused_lloyd_single_step(
...
 
                 accumulate_dot_products(
-                    sample_idx, first_feature_idx, X_t, centroids_window, dot_products, is_last_feature_window, is_last_centroid_window
+                    sample_idx,
+                    first_feature_idx,
+                    X_t,
+                    centroids_window,
+                    dot_products,
+                    is_last_feature_window,
+                    is_last_centroid_window
```
jjerphan, line 245: Nit.
```suggestion
                    is_last_centroid_window,
```

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -41,15 +41,30 @@ def make_lloyd_single_step_fixed_window_kernel(
...
+    # The height of the window on centroids (or, equivalently, the number of features
+    # in the window), and the width (number of centroids in the window), are chosen
+    # such that:
+    # - the width is equal to the `sub_group_size`, which is the mininal size that
+    # optimizes IO patterns by having items in each sub group read contiguous memory
+    # slots, which can optimize IO bandwitch by trigerring coalescence of the read
+    # operations of each item in the sub group. Higher sizes would increase the
+    # pressure on private memory (since the size of the private array
+    # `dot_products` is equal to the width of the window) with no particular gain to
+    # expect.
```
jjerphan, line 53:
```suggestion
    #   - the width is equal to the `sub_group_size`, which is the minimal
    #   size that optimizes IO patterns by having items in each sub group
    #   read contiguous memory slots.  This configuration can optimize IO
    #   bandwidth by triggering coalescence of the read operations of each
    #   item in the sub group. Higher sizes would increase the pressure on
    #   private memory (since the size of the private `dot_products` array
    #   is equal to the width of the window) with no particular gain to expect.
```

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -41,15 +41,30 @@ def make_lloyd_single_step_fixed_window_kernel(
...
+    # - the height is chosen such that the window counts `work_group_size` items.
+    # When a window is cooperatively loaded into shared memory by a work group, each
+    # work item is thus tasked with loading one and only one item.
+    # The number of items is `centroids_window_height * window_n_centroids =
+    # centroids_window_height * sub_group_size`
```
jjerphan, line 58:
```suggestion
    #   - the height is chosen such that the window counts `work_group_size` items.
    #   When a window is cooperatively loaded into shared memory by a work group,
    #   each work item is thus tasked with loading one and only one item.
    #   The number of items is: `centroids_window_height * window_n_centroids`, i.e.
    #   `centroids_window_height * sub_group_size`
```

jjerphan (approved):
LGTM, just nits here.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
ogrisel: I suspect an imbalanced problem between `# fmt: off` / `# fmt: on` but I could not spot it here.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -411,3 +417,39 @@ def test_kmeans_plusplus_dataorder():
...
+
+def test_error_raised_on_invalid_group_sizes():
+    n_samples = 10
+    n_features = 2
+    n_clusters = 2
+    sub_group_size = 64
+    work_group_size = 500  # invalid because is not a multiple of sub_group_size
+    dtype = np.float32
+
+    with pytest.raises(ValueError):
```
ogrisel, line 430: Could you please check the content of the error message as well?

```python
    expected_msg = re.escape(
        "..."  # TODO
    )
    with pytest.raises(ValueError, match=expected_msg):
        ...
```

In addition to checking the correctness of the message it will make the tests easier to read.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
ogrisel: Actually it would be even better to indent the lines within a bullet point to match the column of the first non empty char after the `-` to match Restructured Text / markdown conventions.

ogrisel (approved):
LGTM besides the following details.

Here is the output of the benchmarks on my latop GPU:

- This PR's branch:

```
root@8b981842b196:/io/code/sklearn-numba-dpex# python benchmark/kmeans.py 
/io/code/sklearn-numba-dpex/benchmark/ext_helpers/daal4py.py:8: RuntimeWarning: The current version of scikit-learn ( ==1.2.dev0 ) is too recent to ensure good compatibility with sklearn intelex, who only supports sklearn >=0.22, <1.1 . Use very cautiously, things might not work as expected...
  warnings.warn(
Running Sklearn vanilla lloyd ... done in 7.922611640009563

Running daal4py lloyd CPU ... done in 5.287298984010704

Running Kmeans numba_dpex lloyd GPU with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU ... done in 5.592537177959457
```

- `main` branch:

```
root@8b981842b196:/io/code/sklearn-numba-dpex# python benchmark/kmeans.py 
/io/code/sklearn-numba-dpex/benchmark/ext_helpers/daal4py.py:8: RuntimeWarning: The current version of scikit-learn ( ==1.2.dev0 ) is too recent to ensure good compatibility with sklearn intelex, who only supports sklearn >=0.22, <1.1 . Use very cautiously, things might not work as expected...
  warnings.warn(
Running Sklearn vanilla lloyd ... done in 7.932991191977635

Running daal4py lloyd CPU ... done in 4.825643467018381

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=1) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=1) ... done in 29.45909150701482

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=2) ... done in 5.518104489019606

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=4) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=4) ... done in 5.657601150975097

Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=8) with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU (work_group_size_multiplier=8) ... done in 6.087186309974641
```

so the new automated setting seems optimal compared to the old system.

However, I remember that it used to be the case that the `sklearn_numba_dpex` GPU engine would be (slightly) faster than daal4py CPU engine on this machine. This is no longer the case... But this is not related to this particular PR.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
jjerphan: Done.

fcharras:
> However, I remember that it used to be the case that the `sklearn_numba_dpex` GPU engine would be (slightly) faster than daal4py CPU engine on this machine. This is no longer the case... But this is not related to this particular PR.

Running this test bench on local machines also returns variable results for me that I seem independent of versioning, even when comparing consecutive runs. I wonder if it also can be related to heating, power saving mode, kernel updates, and other processes that could use resources at the same time (like a compositor on the desktop environment).

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
fcharras: Is this diff only formating or is there another change I miss ? why would we enforce 80 characters limits on the inline comments while the character limit for the code is above this threshold (88 IIRC) ?

jjerphan: I should have precised the nature of this suggestion.

This suggestion proposes rephrasing this paragraph (it especially proposes breaking long lines) and does some reformatting of the comment.

A 88 (or even 80) characters' limit is good to me, but I have not checked whether this comment currently complies with such limit.

fcharras:
TY for the review, I'll apply suggestions and merge

PR #50, opened 2022-11-10
Move all input validation steps to engine

fcharras:


fcharras:
Yes it would be more practical to wait for !37. !51 and !49 before resuming here. There's a bit of a bottleneck, the refactoring on this branch is conflicting everything.

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -65,27 +103,208 @@ def init_centroids(self, X):
...
+        # NB: numba_dpex kernels only currently supports inputs with a C memory layout
+        # (see https://github.com/IntelPython/numba-dpex/issues/767) but our KMeans
+        # implementation is hypothetized to be more efficient with the F-memory layout.
+        # As a workaround the kernels work with the transpose of X, X_t, where X_t
+        # is created with a C layout, which results in equivalent memory access
+        # patterns than with a F layout for X.
+        # TODO: when numba_dpex supports inputs with F-layout:
+        # - use X rather than X_t and adapt the codebase (better for readability and
+        # more consistent with sklearn notations)
+        # - test the performances with both layouts and use the best performing layout.
```
fcharras, line 297: NB: I've reformulated this a bit from the previous version

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -358,27 +362,31 @@ def _get_score_with_centers(centers):
             np.mean(scores_engine_kmeans_plusplus),
         ],
         [
-            -1827.2270605322217,
-            -1027.674264781121,
-            -885.1541877428601,
+            -1827.22702,
+            -1027.674243,
+            -865.257397,
```
fcharras, line 367: NB: the engine output changes from the previous version because the `random_state` now go through sklearn random state validation in the `prepare_fit`  step and is transformed as a `np.random.RandomState` object, rather than being passed directly to the driver as an int.

fcharras:
I've rebooted this branch from the current state of https://github.com/soda-inria/sklearn-numba-dpex/pull/49, review can now go on since https://github.com/soda-inria/sklearn-numba-dpex/pull/49 will not undergo significant changes before merge.

Yet again there's a lot of diff but changes are in fact minors, the block of codes for input validation are moved from `drivers.py` to `engine.py`.

I thought this step will make it easier to iterate on for the next work that is planned on interoperability:
- accepting dpnp/dpt arrays as input
- storing the fitted attributes in native types and converting
- parameterizing the output dtype

since instead of having input validation and type conversion written in two files (`engine.py` and `driver.py`) now everything is in the same place `engine.py`.

jjerphan:
To me, PRs whose diff are above ±500 diff really are hard to review.

I find that this is the case of this PR (which has a huge non-trivial diff); is it possible to split it into several smaller PRs? :pray:

fcharras:
This PR is a refacto that moves block of code around.  I don't think that it should be reviewed line by line, the granularity is block of codes. It is opened as a standalone PR to avoid mixing the refacto with actual future changes in logic of input validation within the blocks, I don't think it can be more atomic.

 I tried to explain this in the previous comments here's a more accurate attempt:

- 90% of the changes are blocks of codes that are left unchanged and moved from one file to another or unindented
  - moved: public method from former `KMeansDriver` classes  (that called private methods plus some pre and post processing on raw input data) are moved to the engine
  - unindented: the `KMeansDriver` class abstraction does not make sense anymore and is removed, and the former private functions that were attached become standalone functions, hence losing a level of indentation
  - as a result functions in `drivers.py` only have `dpt.tensor` inputs and outputs. Interfacing with other array libraries is now the responsibility of the engine. 
- changes of interest are the signatures of the functions that were created in the engine and how the tests are impacted
- the code is only moved in one way: from the drivers.py file to the engine.py file

This kind of PR is annoying but isn't it expected for a starting project ? Moreover the test coverage is enough to be confident that there's no regression.

jjerphan (approved):
LGTM. I wish git had a feature to see transfer of unchanged chunks of code between files.

Issue #52, opened 2022-11-14
Remove dependency on pyopencl when dpctl==0.14.0 is out


oleksandr-pavlyk:
As discussed in IntelPython/dpctl#886 `preferred_wor_groups_size_multiple` is a device-specific property of a kernel, rather than the property of `dpctl.SyclDevice`.

`clinfo` retrieves exactly as such. The following Python code uses SPIR-V from simple kernel taken from dpctl example (https://github.com/IntelPython/dpctl/tree/master/examples/pybind11/use_dpctl_sycl_kernel/resource):

```python

import dpctl
import dpctl.program as dpp
import dpctl.tensor._device as dpt_d

double_it_spv = (
    b"\x03\x02#\x07\x00\x04\x01\x00\x0e\x00\x06\x00\x14\x00\x00\x00\x00"
    b"\x00\x00\x00\x11\x00\x02\x00\x04\x00\x00\x00\x11\x00\x02\x00\x05\x00"
    b"\x00\x00\x11\x00\x02\x00\x06\x00\x00\x00\x0b\x00\x05\x00\x01\x00\x00\x00"
    b"OpenCL.std\x00\x00\x0e\x00\x03\x00\x01\x00\x00\x00\x02\x00\x00\x00\x0f"
    b"\x00\x07\x00\x06\x00\x00\x00\t\x00\x00\x00double_it\x00\x00\x00\x05\x00"
    b"\x00\x00\x03\x00\x03\x00\x03\x00\x00\x00p\x8e\x01\x00\x05\x00\x0b\x00\x05"
    b"\x00\x00\x00__spirv_BuiltInGlobalInvocationId\x00\x00\x00\x05\x00\x03\x00"
    b"\n\x00\x00\x00x\x00\x00\x00\x05\x00\x03\x00\x0b\x00\x00\x00y\x00\x00\x00"
    b"\x05\x00\x04\x00\x0c\x00\x00\x00entry\x00\x00\x00\x05\x00\x04\x00\x0e\x00"
    b"\x00\x00call\x00\x00\x00\x00\x05\x00\x05\x00\x0f\x00\x00\x00arrayidx\x00\x00"
    b"\x00\x00\x05\x00\x03\x00\x12\x00\x00\x00mul\x00\x05\x00\x05\x00\x13\x00\x00"
    b"\x00arrayidx1\x00\x00\x00G\x00\r\x00\x05\x00\x00\x00)\x00\x00\x00"
[...]

PR #58, opened 2022-11-17
ENH Up to date dockerfile

fcharras:
First step towards solving https://github.com/soda-inria/sklearn-numba-dpex/issues/52

The use of `dpctl==0.14.0rc2` unlocks early the use of `dpctl.SyclDevice.sub_group_sizes` if you want to check the values on your machines.

I check that all tests passes before pushing and the benchmark is unchanged.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,7 @@
 #
 # TODO: make a custom build with other base images to test compatibility
 # with other OSes.
-ARG BASE_IMAGE_VERSION=latest-f3cc5b196af0d6f69df8c67543f160bb73425b5a
+ARG BASE_IMAGE_VERSION=latest-082cde669552482c541a297e38adce4afee4fd34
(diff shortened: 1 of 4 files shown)


jjerphan (approved):

fcharras:
Mmh it's normal that the test are failing given that the changes require `dpctl>=0.14.0rc2` but maybe we should keep compatibility already since other people are trying to test the package

ogrisel (approved):
I think its fine to depend on 0.14 rc2.

fcharras:
OK then let's merge

Don't forget to pull the new image afterward when testing things, should be available in ~1hr

jjerphan:
It looked like the CI was failing, has the problem it been resolved?

fcharras:
Test pipelines should be rerun when https://github.com/soda-inria/sklearn-numba-dpex/actions/runs/3495827078/jobs/5853027401 is finished, should be green afterward.

PR #62, opened 2022-11-23
FEA: engine accepts dpnp.ndarray and dpt.usm_ndarray objects as input data.

fcharras:
I think I'm happy with the result, the code is a little messy but not more than sklearn's and I think the PR basically succeeds in mimicking almost all aspects of sklearn's UX regarding what inputs are accepted and how it's casted under the hood, and also performance regarding the cautiousness with memory copies, while additionally creating support for `dpnp.ndarray` and `dpt.usm_ndarray` inputs, and recycling the most possible code that already exists in sklearn.

The PR still needs some polishing, and a few tests with `dpnp`/`dpt` tensors.

I think one (minor) difference is that, while sklearn try to convert numpy arrays with `object` dtype to `float64`, our engine will error out in this case because `dpt.asarray` refuses object dtype as input. Those kind of edge cases are complicated to unifiy because different array libraries can have different choices and I think it's fine we let it fail in this case.

### DIFF
--- a/benchmark/ext_helpers/daal4py.py
+++ b/benchmark/ext_helpers/daal4py.py
@@ -1,33 +1,38 @@
 import warnings
 import sklearn
 
+
 # HACK: daal4py will fail to import with too recent versions of sklearn because
 # of this missing attribute. Let's pretend it still exists.
-
-if not hasattr(sklearn.neighbors._base, "_check_weights"):
-    warnings.warn(
-        f"The current version of scikit-learn ( =={sklearn.__version__} ) is too "
-        "recent to ensure good compatibility with sklearn intelex, who only supports "
-        "sklearn >=0.22, <1.1 . Use very cautiously, things might not work as "
-        "expected...",
-        RuntimeWarning,
-    )
-    try:
+try:
+    if not hasattr(sklearn.neighbors._base, "_check_weights"):
+        warnings.warn(
+            f"The current version of scikit-learn ( =={sklearn.__version__} ) is too "
+            "recent to ensure good compatibility with sklearn intelex, who only supports "
+            "sklearn >=0.22, <1.1 . Use very cautiously, things might not work as "
+            "expected...",
+            RuntimeWarning,
+        )
+        _daal4py_kmeans_compat_mode = True
         sklearn.neighbors._base._check_weights = None
-        from daal4py.sklearn.cluster._k_means_0_23 import KMeans
-    finally:
+    from daal4py.sklearn.cluster._k_means_0_23 import (
+        KMeans,
+        _daal4py_compute_starting_centroids,
+        getFPType,
+    )
+finally:
+    if _daal4py_kmeans_compat_mode:
         del sklearn.neighbors._base._check_weights
-else:
-    from daal4py.sklearn.cluster._k_means_0_23 import KMeans
+
 
 from sklearn.exceptions import NotSupportedByEngineError
-from sklearn_numba_dpex.kmeans.engine import KMeansEngine
+from sklearn.cluster._kmeans import KMeansCythonEngine
 
 
 # TODO: instead of relying on monkey patching the default engine, find a way to
 # register a distinct entry point that can load a distinct engine outside of setup.py
 # (impossible ?)
-class DAAL4PYEngine(KMeansEngine):
+class DAAL4PYEngine(KMeansCythonEngine):
     def prepare_fit(self, X, y=None, sample_weight=None):
         if sample_weight is not None and any(sample_weight != sample_weight[0]):
             raise NotSupportedByEngineError(
@@ -37,7 +42,15 @@ class DAAL4PYEngine(KMeansEngine):
         return super().prepare_fit(X, y, sample_weight)
 
     def init_centroids(self, X):
-        return super(KMeansEngine, self).init_centroids(X)
+        _, centroids = _daal4py_compute_starting_centroids(
+            X,
+            getFPType(X),
+            self.estimator.n_clusters,
+            self.estimator.init,
+            self.estimator.verbose,
+            self.random_state,
+        )
+        return centroids
 
     def kmeans_single(self, X, sample_weight, centers_init):
 
--- a/benchmark/ext_helpers/daal4py.py
+++ b/benchmark/ext_helpers/daal4py.py
@@ -29,6 +29,7 @@ finally:
 from sklearn.exceptions import NotSupportedByEngineError
 from sklearn.cluster._kmeans import KMeansCythonEngine
 
+
 # TODO: instead of relying on monkey patching the default engine, find a way to
 # register a distinct entry point that can load a distinct engine outside of setup.py
 # (impossible ?)
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -163,6 +163,93 @@ def make_half_l2_norm_2d_axis0_kernel(size0, size1, work_group_size, dtype):
 
 
 @lru_cache
+def make_sum_reduction_2d_axis1_kernel(
+    size0, size1, work_group_size, device, dtype, fused_unary_func=None
+):
+    """Implement data_2d.sum(axis=1) or data_1d.sum()
+
+    numba_dpex does not provide tools such as `cuda.reduce` so we implement from scratch
+    a reduction strategy. The strategy relies on the commutativity of the operation used
+    for the reduction, thus allowing to reduce the input in any order.
+
+    The strategy consists in performing local reductions in each work group using local
+    memory where each work item combine two values, thus halving the number of values,
+    and the number of active work items. At each iteration the work items are discarded
+    in a bracket manner. The work items with the greatest ids are discarded first, and
+    we rely on the fact that the remaining work items are adjacents to optimize the RW
+    operations.
+
+    Once the reduction is done in a work group the result is written in global memory,
+    thus creating an intermediary result whose size is divided by
+    `2 * work_group_size`. This is repeated as many time as needed until only one value
+    remains in global memory.
+
+    If fused_unary_func is not None, it will be applied element-wise before summing.
+    It must be a function that will be interpreted as a dpex.func and is subject to the
+    same rules. It is expected to take one scalar argument and returning one scalar
+    value. lambda functions are advised against since the cache will not work with
+    lambda functions. sklearn_numba_dpex.common._utils expose some pre-defined
+    `fused_unary_funcs`.
+
+    Notes
+    -----
+    `work_group_size` is assumed to be a power of 2.
+
+    if `size1` is None then the kernel expects 1d tensor inputs. If `size1` is not None
+    then the expected shape of input tensors is `(size0, size1)`, and the reduction
+    operation is equivalent to input.sum(axis=1). In this case, the kernel is a good
+    choice if `size1` >> `preferred_work_group_size_multiple`, and if `size0` ranges in
+    the same order of magnitude than `preferred_work_group_size_multiple`. If not,
+    other reduction implementations might give better performances.
+    """
+    check_power_of_2(work_group_size)
+    n_rows = size0 if size1 is not None else None
+    sum_axis_size = size0 if n_rows is None else size1
+
+    # fused_unary_func is applied elementwise during the first pass on data, in the
+    # first kernel execution only.
+    fused_func_kernel = _make_partial_sum_reduction_2d_axis1_kernel(
+        n_rows, work_group_size, fused_unary_func, dtype
+    )
+    # subsequent kernel calls only sum the data.
+    nofunc_kernel = _make_partial_sum_reduction_2d_axis1_kernel(
+        n_rows, work_group_size, fused_unary_func=None, dtype=dtype
+    )
+
+    # As many partial reductions as necessary are chained until only one element
+    # remains.
+    kernels_and_empty_tensors_pairs = []
+    n_groups = sum_axis_size
+    # TODO: at some point, the cost of scheduling the kernel is more than the cost of
+    # running the reduction iteration. At this point the loop should stop and then a
+    # single work item should iterates one time on the remaining values to finish the
+    # reduction.
+    kernel = fused_func_kernel
+    while n_groups > 1:
+        n_groups = math.ceil(n_groups / (2 * work_group_size))
+        global_size = n_groups * work_group_size
+        kernel = kernel[global_size, work_group_size]
+        result_shape = n_groups if n_rows is None else (n_rows, n_groups)
+        # NB: here memory for partial results is allocated ahead of time and will only
+        # be garbage collected when the instance of `sum_reduction` is garbage
+        # collected. Thus it can be more efficient to re-use a same instance of
+        # `sum_reduction` (e.g within iterations of a loop) since it avoid deallocation
+        # and reallocation every time.
+        result = dpt.empty(result_shape, dtype=dtype, device=device)
+        kernels_and_empty_tensors_pairs.append((kernel, result))
+        kernel = nofunc_kernel
+
+    def sum_reduction(summands):
+        # TODO: manually dispatch the kernels with a SyclQueue
+        for kernel, result in kernels_and_empty_tensors_pairs:
+            kernel(summands, result)
+            summands = result
+        return result
+
+    return sum_reduction
+
+
+@lru_cache
 def _make_partial_sum_reduction_2d_axis1_kernel(
     n_rows, work_group_size, fused_unary_func, dtype
 ):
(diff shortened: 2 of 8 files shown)


benchmark/ext_helpers/daal4py.py
```diff
@@ -37,7 +42,15 @@ def prepare_fit(self, X, y=None, sample_weight=None):
         return super().prepare_fit(X, y, sample_weight)
 
     def init_centroids(self, X):
-        return super(KMeansEngine, self).init_centroids(X)
```
fcharras, line 40: This is a very important bugfix, I will open in a separate PR to fasttrack.

Basically, currently on main `k-means++` benchmark with the daal actually runs sklearn's. After fixing, we can see that the daal is more that 10 times faster than sklearn at kmeans++.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -304,6 +316,53 @@ def partial_sum_reduction(
         if first_work_id:
             add_first_cols(local_values, result, group_id)
 
+    return partial_sum_reduction
+
+
+@lru_cache
+def make_sum_reduction_2d_axis1_kernel(
+    size0, size1, work_group_size, device, dtype, fused_unary_func=None
```
fcharras, line 324: The changes to this sum are the addition of this arg `fused_unary_func` which enable fusing a unary ops on the elements of the input before summing. It's used in this PR to compute variance.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -364,7 +366,7 @@ def _get_score_with_centers(centers):
         [
             -1827.22702,
             -1027.674243,
-            -865.257397,
+            -865.257501,
```
fcharras, line 369: The difference in those values comes from numerical instability after including the X_mean removal step.

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -156,155 +201,160 @@ def kmeans_single(self, X, sample_weight, centers_init):
...
+    # remove those hacks.
+    with sklearn.config_context(
+        array_api_dispatch=True,
+        assume_finite=True  # workaround 1: disable force_all_finite
+        # workaround 2: monkey patch get_namespace and _asarray_with_order to force
+        # dpctl.tensor array namespace
+    ), override_attr_context(
+        sklearn_validation,
+        get_namespace=_get_namespace,
+        _asarray_with_order=_asarray_with_order,
```
fcharras, line 358: Using monkey patching here enables a very efficient workflow as a whole. I wonder if rather than having those 2 functions limited to array api specs, we should also consider pluginying it.

benchmark/ext_helpers/daal4py.py
ogrisel: It would be very interesting to repriorize the optimization of kmeans++, both algorithmically (via subsampling as suggested by @jeremiedbb elsewhere) and with efficient Cython code for the computation of the argmin of distances if it's not already the case.

fcharras:
Out of WIP

sklearn_numba_dpex/testing/config.py
```diff
@@ -31,14 +31,15 @@ def override_attr_context(obj, **attrs):
...
+        for attr_name, attr_value in attrs.items():
+            # raise AttributeError if obj does not have the attribute attr_name
+            attrs_before[attr_name] = getattr(obj, attr_name)
+            setattr(obj, attr_name, attr_value)
+
+        yield
+
+    finally:
+        for attr_name, attr_value in attrs_before.items():
+            setattr(obj, attr_name, attr_value)
```
jjerphan, line 45: How about this?
```suggestion
      attrs_before = dict()
      for attr_name, attr_value in attrs.items():
          attribute = getattr(obj, attr_name, None)
          if attribute is not None:
              # Only replace the value of the `attr_name` does exist.
              attrs_before[attr_name] = attribute
              setattr(obj, attr_name, attr_value)

      yield

      for attr_name, attr_value in attrs_before.items():
          setattr(obj, attr_name, attr_value)
```

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -336,8 +342,10 @@ def _get_score_with_centers(centers):
 
         kmeans.set_params(random_state=random_state)
         engine = KMeansEngine(kmeans)
-        engine.prepare_fit(X)
-        engine_kmeans_plusplus_centers = engine.init_centroids(X)
+        X_prepared, *_ = engine.prepare_fit(X)
+        engine_kmeans_plusplus_centers = engine.init_centroids(X_prepared)
+        engine_kmeans_plusplus_centers = dpt.asnumpy(engine_kmeans_plusplus_centers.T)
```
jjerphan, line 347: Do we need to `_t`-suffix one of them?

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -364,29 +372,36 @@ def _get_score_with_centers(centers):
...
-        random_state=random_state,
+        init="k-means++", n_clusters=n_clusters_sklearn_test, random_state=random_state
     )
     engine = KMeansEngine(estimator)
-    engine.prepare_fit(X, sample_weight=sample_weight)
-    centers, indices = engine._kmeans_plusplus(X)
+    X_prepared, *_ = engine.prepare_fit(X, sample_weight=sample_weight)
+
+    centers, indices = engine._kmeans_plusplus(X_prepared)
+    centers = dpt.asnumpy(centers.T)
```
jjerphan, line 402: Similarly, do we need to `_t`-suffix one of them?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -172,12 +171,16 @@ def make_sum_reduction_2d_axis1_kernel(size0, size1, work_group_size, device, dt
     minus_one_idx = np.int64(-1)
     two_as_a_long = np.int64(2)
 
-    is_1d = size1 is None
+    if fused_unary_func is None:
```
jjerphan, line 174: Can you document `fused_unary_func`, please?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -401,6 +402,88 @@ def _relocate_empty_clusters(
...
+        X_t.shape[1],
+        device.max_work_group_size,
+        device=device,
+        dtype=compute_dtype,
+    )
+
+    X_mean = (sum_axis1_kernel(X_t) / compute_dtype(n_samples))[:, 0]
+
+    if (X_mean == 0).astype(int).sum() == len(X_mean):
+        X_mean = None
```
jjerphan, line 423: Can you document why `None` is conventionally used in this case?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -401,6 +402,88 @@ def _relocate_empty_clusters(
...
+                n_clusters,
+                ops=_minus,
+                work_group_size=max_work_group_size,
+            )
+            broadcast_init_minus_X_mean(init, X_mean)
+
+    variance_kernel = make_sum_reduction_2d_axis1_kernel(
+        n_features * n_samples,
+        None,
+        max_work_group_size,
```
jjerphan, line 448:
```suggestion
        size0=n_features * n_samples,
        size1=None,
        work_group_size=max_work_group_size,
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -401,6 +402,88 @@ def _relocate_empty_clusters(
...
+    )
+    broadcast_X_plus_X_mean(X_t, X_mean)
+
+
+def _minus(x, y):
+    return x - y
+
+
+def _plus(x, y):
+    return x + y
```
jjerphan, line 484: Can you group and document those helpers function for `ops`?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -663,4 +746,4 @@ def kmeans_plusplus(
         centers_t[:, c] = X_t[:, center_index]
         center_indices[c] = center_index
 
-    return centers_t.T, center_indices
+    return centers_t, center_indices
```
jjerphan, line 749: What is the reason for this change (i.e. not transposing anymore)?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -100,6 +100,34 @@ def broadcast_division(dividend_array, divisor_vector):
...
+    value. lambda functions are advised against since the cache will not work with lamda
+    functions."""
+
+    global_size = math.ceil(size1 / work_group_size) * work_group_size
+    ops = dpex.func(ops)
+
+    # NB: inplace. # Optimized for C-contiguous array and for
+    # size1 >> preferred_work_group_size_multiple
+    @dpex.kernel
+    def broadcast_ops(left_operand_array, right_operand_vector):
```
jjerphan, line 117: I think it's worth indicating that the left operand is modified in place and that the right one isn't modified at all.

jjerphan (commented):
Thank you, @fcharras. Here is a first pass.

sklearn_numba_dpex/common/kernels.py
fcharras: It's documented in the public function that follows. For readability it would be better I think to swap the definitions of those, but it would have increased the diff masked the true diff.

sklearn_numba_dpex/kmeans/drivers.py
fcharras: "drivers" in drivers.py file now take and return the raw-est input and outputs. Pre and post processing regarding transposing, type, dtype, etc, that is required upstream or downstream is now located in the engine. I think those are better concepts for readability and maintaining.

sklearn_numba_dpex/testing/config.py
fcharras: The `try/finally` is not only about the `AttributeError` that was documented, but about any error that can be raised after `yield` and before exiting the context. I thought initially that before exiting the code after `yield` is always executed, in the same way it's true for  `__exit__` when expliciting context managers, but in case of such errors it seems not to be true, and caused failing tests (in the sklearn pipeline). Adding the `try/finally` fixes that.

About only replacing the value if `attr_name` does exist, i'd prefer not, raising an `AttributeError` is a safeguard against typos, or unseen changes in input objects, when using this context manager. Its only intended uses are for replacing attributes that exist and I like it to fail otherwise.

fcharras:
Pushed your suggestions, also amended some changes to enable compute follow data. (the preferred device for compute is the device that stores de data if data is already on-device)

sklearn_numba_dpex/common/kernels.py
```diff
@@ -304,6 +310,63 @@ def partial_sum_reduction(
...
+    sum_axis_size = size0 if n_rows is None else size1
+
+    # fused_unary_func is applied elementwise during the first pass on data, in the
+    # first kernel execution only.
+    fused_func_kernel = _make_partial_sum_reduction_2d_axis1_kernel(
+        n_rows, work_group_size, fused_unary_func, dtype
+    )
+    # subsequent kernel calls only sum the data.
+    nofunc_kernel = _make_partial_sum_reduction_2d_axis1_kernel(
+        n_rows, work_group_size, None, dtype
```
jjerphan, line 367:
```suggestion
        n_rows, work_group_size, fused_unary_func=None, dtype=dtype
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -401,6 +404,80 @@ def _relocate_empty_clusters(
...
+        X_t.shape[0],
+        X_t.shape[1],
+        device.max_work_group_size,
+        device=device,
+        dtype=compute_dtype,
+    )
+
+    X_mean = (sum_axis1_kernel(X_t) / compute_dtype(n_samples))[:, 0]
+
+    if (X_mean == 0).astype(int).sum() == len(X_mean):
```
jjerphan, line 424:
```suggestion
    X_mean_is_zeroed = (X_mean == 0).astype(int).sum() == len(X_mean)

    if X_mean_is_zeroed:
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -401,6 +404,80 @@ def _relocate_empty_clusters(
     )
 
 
+def prepare_data_for_lloyd(X_t, init, tol, copy_x):
```
jjerphan, line 407: Can you document `prepare_data_for_lloyd` indicating that this is centering `X` for numerical stability?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -401,6 +404,80 @@ def _relocate_empty_clusters(
...
+        dtype=compute_dtype,
+        fused_unary_func=_square,
+    )
+    variance = variance_kernel(dpt.reshape(X_t, -1)) / n_features
+    tol = variance * tol
+
+    return X_t, X_mean, init, tol
+
+
+def restore_data_after_lloyd(X_t, X_mean):
```
jjerphan, line 465: Similarly, can you document `restore_data_after_lloyd` indicating that this is translating `X` to its original position?

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -47,101 +63,134 @@ class KMeansEngine(KMeansCythonEngine):
...
+        elif isinstance(init, str) and init == "k-means++":
+            centers_t, _ = self._kmeans_plusplus(X)
+
+        elif callable(init):
+            centers = init(X, self.estimator.n_clusters, random_state=self.random_state)
+            centers_t = self._check_init(centers, X)
+
+        else:
+            # NB: sampling without replacement must be executed sequentially so
+            # it's better done on CPU
```
jjerphan, line 161: :+1:

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -47,101 +63,134 @@ class KMeansEngine(KMeansCythonEngine):
...
+            centers_idx = self.random_state.choice(
+                X.shape[0], size=n_clusters, replace=False
+            )
+            # Poor man's fancy indexing
+            # TODO: write a kernel ? or replace with better equivalent when available ?
+            centers_t = dpt.concat(
+                [dpt.expand_dims(X[center_idx], axes=1) for center_idx in centers_idx],
+                axis=1,
+            )
```
jjerphan, line 170: Should we open an issue on `dpctl`? What do you think?

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -47,101 +63,134 @@ class KMeansEngine(KMeansCythonEngine):
...
         )
+        return centers_t, center_indices
+
+    def kmeans_single(self, X, sample_weight, centers_init_t):
+        # ???: using `.all()` often segfaults
+        # TODO: minimal reproducer and issue at dpnp
+        # or write a kernel ?
+        use_uniform_weights = (sample_weight == sample_weight[0]).astype(
+            int
+        ).sum() == len(sample_weight)
```
jjerphan, line 188:
```suggestion
        use_uniform_weights = (
            (sample_weight == sample_weight[0]).astype(int).sum()
            == len(sample_weight)
        )
```

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -156,155 +205,170 @@ def kmeans_single(self, X, sample_weight, centers_init):
...
         )
 
+    def prepare_prediction(self, X, sample_weight):
+        X = self._validate_data(X, reset=False)
+        sample_weight = self._check_sample_weight(sample_weight, X)
+        return X, sample_weight
+
     def get_labels(self, X, sample_weight):
-        labels, _ = self._get_labels_inertia(X, with_inertia=False)
+        # TODO: sample_weight actually not used for get_labels. Fix in sklearn ?
```
jjerphan, line 223: Is there an issue already open for it?

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -156,155 +205,170 @@ def kmeans_single(self, X, sample_weight, centers_init):
...
-            text = (
-                f"KMeans is set to compute with type {compute_dtype} but this type is "
-                f"not supported by the device {self.device.name}. The computations "
-                f"will default back to float32 type."
-            )
-            compute_dtype = np.float32
-
+            sample_weight = dpt.ones(n_samples, dtype=dtype, device=device)
+        elif isinstance(sample_weight, numbers.Number):
+            sample_weight = dpt.full(n_samples, 1, dtype=dtype, device=device)
```
jjerphan, line 306: What are the reasons not to have only one branch here?

jjerphan (approved):
LGTM up to a few suggestions and questions.

Thank you, @fcharras!

sklearn_numba_dpex/kmeans/engine.py
fcharras: A mistake! the latter line should read:

```suggestion
            sample_weight = dpt.ones(n_samples, dtype=dtype, device=device)
        elif isinstance(sample_weight, numbers.Number):
            sample_weight = dpt.full(n_samples, sample_weight, dtype=dtype, device=device)
```

sklearn_numba_dpex/kmeans/engine.py
```diff
@@ -47,101 +62,134 @@ class KMeansEngine(KMeansCythonEngine):
 
     """
 
+    # This class attribute can alter globally the attributes `device` and `order` of
+    # future instances. It is only used for testing purposes, using
+    # `sklearn_numba_dpex.testing.config.override_attr_context` context, for instance
+    # in the benchmark script.
+    # For normal usage, the compute will follow the __compute_follow_data__ principle.
```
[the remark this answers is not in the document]
fcharras: Yes, let me replace with `*compute follow data*` which would be more correct makrdown.

The concept is anchored in dpctl/numba_dpex desiogn

- in dpctl [1](https://intelpython.github.io/dpctl/latest/docfiles/dpctl/dpctl.utils_pyapi.html) 2

- in numba dpex [1](https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/selecting_device.html#defining-the-execution-queue-for-a-kernel-function) [2](https://github.com/IntelPython/numba-dpex/issues/816#issuecomment-1323080669)

sklearn_numba_dpex/kmeans/drivers.py
fcharras: :heavy_check_mark: mostly reported [sklearn docstring](https://github.com/scikit-learn/scikit-learn/blob/17df37aee/sklearn/cluster/_kmeans.py#L332)

sklearn_numba_dpex/kmeans/engine.py
fcharras: `dpctl` implements the Array API and the Array API doesn't list such tools. Best things to have would be `take` but here are the list of supported functions [1](https://data-apis.org/array-api/latest/API_specification/manipulation_functions.html) [2](https://data-apis.org/array-api/latest/API_specification/indexing.html)

I'm not aware enough of where the array api is going to say what would be the best action here :thinking:

sklearn_numba_dpex/kmeans/engine.py
fcharras: Done in https://github.com/scikit-learn/scikit-learn/issues/25066

sklearn_numba_dpex/kmeans/engine.py
fcharras: Actually `take` has been added recently to the Array APi, [dpctl issue](https://github.com/IntelPython/dpctl/issues/1003)

fcharras:
Thanks for the review @jjerphan I've pushed your suggestions and answered other comments. I'll go ahead and merge when the pipeline is green.

Issue #61, opened 2022-11-23
Issue in evaluation of sklearn-numba-dpex with kmeans++


fcharras:
Currently on main the benchmark with sklearn-numba-dpex with init set to k-means++ is actually running vanilla sklearn kmeans++. Will issue a fix soon.

Pinging [Vika-F](https://github.com/Vika-F) in case you try to run those benchmarks too.

PR #63, opened 2022-11-24
fix kmeansplusplus daal benchmark

fcharras:
Fixes https://github.com/soda-inria/sklearn-numba-dpex/issues/61

### DIFF
--- a/benchmark/ext_helpers/daal4py.py
+++ b/benchmark/ext_helpers/daal4py.py
@@ -1,24 +1,30 @@
 import warnings
 import sklearn
 
+
 # HACK: daal4py will fail to import with too recent versions of sklearn because
 # of this missing attribute. Let's pretend it still exists.
-
-if not hasattr(sklearn.neighbors._base, "_check_weights"):
-    warnings.warn(
-        f"The current version of scikit-learn ( =={sklearn.__version__} ) is too "
-        "recent to ensure good compatibility with sklearn intelex, who only supports "
-        "sklearn >=0.22, <1.1 . Use very cautiously, things might not work as "
-        "expected...",
-        RuntimeWarning,
-    )
-    try:
+try:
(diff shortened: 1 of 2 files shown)


fcharras:
It's also a bit cleaner that what was proposed in https://github.com/soda-inria/sklearn-numba-dpex/pull/62 since the engine accesses low level functions directly while still ensuring appropriate device interfacing.

Also, found out that daal4py does not implements gpu kmean++, only cpy.

`n_local_trials` parameter for `daal4py` is consistent with the default in `sklearn` and ours.

ogrisel (approved):
LGTM but I wonder why we don't just use:

```python
import sklearnex
sklearnex.patch_sklearn()
```

and then:

```python
sklearnex.unpatch_sklearn()
```

in the benchmark. This way we wouldn't have to maintain our own way to monkeypatch sklearn just for the benchmark. Agreed, we would still need the hack for the missing attribute.

I am find with fixing the current "ext_helper" engine if you prefer.

Can you copy and paste the output of the benchmark script recorded on your machine to a comment of this PR out of curiosity?

fcharras:
It didn't come to mind that the hacky fix in our `daal4py` module would be enough to enable `patch` and `unpatch`, but it's true that `patch` works ! :astonished: I'm still wondering about what else could be broken though. I think I prefer the current way.

fcharras:
Running on my machine with defaults (which includes `kmeans++`):

```
Running Sklearn vanilla lloyd ... done in 3.0 s

Running daal4py lloyd CPU ... done in 0.6 s

Running Kmeans numba_dpex lloyd GPU with parameters sample_weight=unary n_clusters=127 data_shape=(263256, 14) max_iter=100 tol=0 ...
Running Kmeans numba_dpex lloyd GPU ... done in 1.7 s
```

PR #68, opened 2022-12-02
ENH: Refresh the guide for  installation using conda

fcharras:

PR #69, opened 2022-12-05
FIX: Prepare for dpnp 0.11.0

fcharras:
This PR is a prerequisite for https://github.com/soda-inria/sklearn-numba-dpex/pull/68

Before  `dpnp==0.11.0` (NB: our current image uses `0.11.0rc1`) some calls to `dpnp` could silently fallback on `numpy`. [From `0.11.0` on](https://github.com/IntelPython/dpnp/pull/1201) the silent fallback is removed and an error is raised instead (the fallback mode can be reactivated with an environment variable `DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK`).

That breaks our code at several locations:
- direct call to `take` triggers a fallback
- broadcasting binary operations [triggers fallbacks](https://github.com/IntelPython/dpnp/issues/1238) (!!) 

I'm worried that silent fallbacks on numpy when running on gpu could have dramatic effects on performance if it could trigger memory transfers. Rather than setting this environment variable to reactivate the fallbacks I'd prefer fixing the code so that fallbacks are not needed anymore.

The fixes do not affect parts of the code that are critical for performance so the benchmarks are not affected (luckily or not I wonder :grin: )

### DIFF
--- a/sklearn_numba_dpex/common/_utils.py
+++ b/sklearn_numba_dpex/common/_utils.py
@@ -17,3 +17,7 @@ def _minus(x, y):
 
 def _plus(x, y):
     return x + y
+
+
+def _divide(x, y):
+    return x / y
--- a/sklearn_numba_dpex/kmeans/drivers.py
+++ b/sklearn_numba_dpex/kmeans/drivers.py
@@ -6,11 +6,12 @@ import dpctl.tensor as dpt
 import dpnp
 import numpy as np
 
-from sklearn_numba_dpex.common._utils import _minus, _plus, _square
+from sklearn_numba_dpex.common._utils import _divide, _minus, _plus, _square
 from sklearn_numba_dpex.common.kernels import (
     make_argmin_reduction_1d_kernel,
     make_broadcast_division_1d_2d_axis0_kernel,
     make_broadcast_ops_1d_2d_axis1_kernel,
+    make_elementwise_binary_ops_1d_kernel,
     make_half_l2_norm_2d_axis0_kernel,
     make_initialize_to_zeros_2d_kernel,
     make_initialize_to_zeros_3d_kernel,
@@ -266,6 +267,7 @@ def lloyd(
         compute_centroid_shifts_kernel(centroids_t, new_centroids_t, centroid_shifts)
 
         centroid_shifts_sum, *_ = reduce_centroid_shifts_kernel(centroid_shifts)
+        centroid_shifts_sum = compute_dtype(centroid_shifts_sum)
 
         # ???: unlike sklearn, sklearn_intelex checks that pseudo_inertia decreases
         # and keep an additional copy of centroids that is updated only if the
@@ -403,7 +405,7 @@ def _relocate_empty_clusters(
     )
 
 
-def prepare_data_for_lloyd(X_t, init, tol, copy_x):
+def prepare_data_for_lloyd(X_t, init, tol, sample_weight, copy_x):
     """It can be more numerically accurate to center the data first. If copy_x is True,
     then the original data is not modified. If False, the original data is modified,
     and put back later on (see `restore_data_after_lloyd`), but small numerical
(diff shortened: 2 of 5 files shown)


jjerphan:
Thanks for the PR. If I understand correctly, those changes are only the most minimal needed ones not to raise an error?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -455,18 +476,36 @@ def prepare_data_for_lloyd(X_t, init, tol, copy_x):
...
         dtype=compute_dtype,
         fused_unary_func=_square,
     )
-    variance = variance_kernel(dpt.reshape(X_t, -1)) / n_features
-    tol = variance * tol
 
-    return X_t, X_mean, init, tol
+    variance = variance_kernel(dpt.reshape(X_t, -1))
+
+    tol = (float(variance[0]) / n_features) * tol
```
jjerphan, line 492:
```suggestion
    variance = variance_kernel(dpt.reshape(X_t, -1))[0]

    tol = (float(variance) / n_features) * tol
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -740,8 +779,10 @@ def kmeans_plusplus(
     candidate_ids = dpt.empty(sh=(n_local_trials,), dtype=np.int32, device=device)
 
     # Pick first center randomly
-    starting_center_id = np.int32(get_random_raw(random_state) % n_samples)
-    center_indices[0] = starting_center_id[0]
+    starting_center_id = dpt.asnumpy(get_random_raw(random_state))[0] % np.uint64(
+        n_samples
+    )
+    center_indices[0] = np.int32(starting_center_id)
```
jjerphan, line 785: Shouldn't it be something like this, here?
```suggestion
    center_indices[0] = np.int32(starting_center_id[0])
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -536,7 +575,7 @@ def get_nb_distinct_clusters(labels, n_clusters):
 
     get_nb_distinct_clusters_kernel(labels, clusters_seen, nb_distinct_clusters)
 
-    return nb_distinct_clusters[0]
+    return nb_distinct_clusters.dtype.type(nb_distinct_clusters[0])
```
jjerphan, line 578: What is this doing here?

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -425,8 +427,27 @@ def prepare_data_for_lloyd(X_t, init, tol, copy_x):
...
+        work_group_size=max_work_group_size,
+        device=device,
+        dtype=compute_dtype,
+    )
+
+    X_mean = sum_axis1_kernel(X_t)[:, 0]
+
+    divisor = dpt.full(
+        sh=(1,), fill_value=compute_dtype(n_samples), dtype=compute_dtype, device=device
+    )
```
jjerphan, line 446: Can divisor be null?

sklearn_numba_dpex/common/_utils.py
```diff
@@ -17,3 +17,7 @@ def _minus(x, y):
 
 def _plus(x, y):
     return x + y
+
+
+def _divide(x, y):
+    return x / y
```
jjerphan, line 23: What is our convention for division by zero?

Should we use this?
```suggestion
    return 0. if y == 0. else x / y
```

sklearn_numba_dpex/kmeans/drivers.py
fcharras: That would mean there's 0 samples, that is not possible since the input validation pipeline enforces at least one sample. So, no.

sklearn_numba_dpex/common/_utils.py
fcharras: I thought it's ok to use whatever numba_dpex decides. 

What is the usual convention ? why return 0. for x / 0 rather than +/-inf ?

fcharras:
TY for the review @ogrisel @jjerphan , comments are addressed (either last commits or answers)

sklearn_numba_dpex/kmeans/drivers.py
jjerphan: OK, let's break it in two pieces if this LGTY. :)

sklearn_numba_dpex/common/_utils.py
jjerphan: Returning 0 allow not overshooting other results when addition is used (which is not the case when using ±∞).

Either way, I am mostly happy if the behavior is clear and it not fatal (which might already be the case: I haven't checked).

Anyway, I do not want to block this PR from advancing and let you decide the best decision to take for this case.

jjerphan (approved):
LGTM!

sklearn_numba_dpex/common/_utils.py
fcharras: Anyway currently there aren't usecases with possible division by zero.

Let's decide this when it's an issue with future work and we'll align conventions with existing scikit-learn behavior so we can pass scikit-learn unit tests.

Issue #54, opened 2022-11-15
Possible typo in `kernels.py:sum_reduction` function


oleksandr-pavlyk:
https://github.com/soda-inria/sklearn-numba-dpex/blob/main/sklearn_numba_dpex/common/kernels.py#L221-L225

The function reads

```python
def sum_reduction(summands):
    for kernel, result in kernels_and_empty_tensors_pairs:
        kernel(summands, result) 
        summands = result
    return result
```

The returned variable `result` is outside of its scope of definition (local variable defined in the scope of the loop). 

I believe the last line should be `return summands` instead.

fcharras:
I can change this in a PR if it's more readable, but in fact in python the scope of the variables defined in a loop is the same scope that other local variables. 

e.g you can do

```python
for i in range(10):
    pass
print(i)
```

output will be `9`.

BTW we just merged unit tests for the reduction kernels and added a reduction for 2d tensors on axis 1 :grin: 

(See https://github.com/soda-inria/sklearn-numba-dpex/pull/46 for some thinking on using custom kernels vs dpnp functions)

oleksandr-pavlyk:
Thanks for highlighting #46 ! 

I think `return summands` would read better indeed, but its just a suggestion.

PR #72, opened 2022-12-07
MNT Overall cleaning and readability

fcharras:
A minor PR with several leftover small tasks I had on a list:

- [x] add -v tag to pytest run
- [x] remove dependency to pytest (bugfix)
    - main public classes don't work if pytest is not installed because they import `override_attr_context` which is defined in a module that imports pytest, without it being necessary
- [x] add OUT inline comments in kernel calls
- [x] sums should return summands rather than result ?
   Closes https://github.com/soda-inria/sklearn-numba-dpex/issues/54


jjerphan (approved):
LGTM. This reads much more conveniently now.

PR #79, opened 2022-12-23
ENH Enhances and refactors sum kernel

fcharras:
I noticed that our sum reduction kernel was awkwardly written. The enhancement to size1 > 1 was done wrong. Each work group consumes shared memory equal to work_group_size * size1 but that is unjustified and will even cause shared memory overflow when size1 is too big.

The root cause it that there's a loop on `n_rows` when size1 > 1, but it was only necessary to repeatedly apply the same kernel to the other rows. That enables a better allocation of shared memory (constant w.r.t size1) and also reads much better overall since it doesn't require to abstract the operator anymore.

PR #80, opened 2022-12-26
FEAT implement sum over axis0

fcharras:
Follows https://github.com/soda-inria/sklearn-numba-dpex/pull/79

Implements sum over axis 0

It seems that this kernel will be necessary for implementation of topK.

### DIFF
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -27,6 +27,7 @@ import numpy as np
 from sklearn_numba_dpex.common._utils import (
     _check_max_work_group_size,
     check_power_of_2,
+    get_maximum_power_of_2_smaller_than,
 )
 
 zero_idx = np.int64(0)
@@ -232,16 +233,18 @@ def make_sum_reduction_2d_kernel(
 
     Once the reduction is done in a work group the result is written in global memory,
     thus creating an intermediary result whose size is divided by
-    `2 * work_group_size` if axis=1, or by `2 * work_group_size // sub_group_size` if\
+    `2 * work_group_size` if axis=1, or by `2 * work_group_size // sub_group_size` if
     axis=0. This is repeated as many times as needed until only one value remains in
     global memory.
 
     If `fused_elementwise_func` is not None, it will be applied element-wise before
     summing. It must be a function that will be interpreted as a dpex.func and is
-    subject to the same rules. It is expected to take one scalar argument and returning
-    one scalar value. lambda functions are advised against since the cache will not
-    work with lambda functions. sklearn_numba_dpex.common._utils expose some
-    pre-defined `fused_elementwise_func`s.
+    subject to the same rules ( see
+    https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html # noqa
+    ). It is expected to take one scalar argument and returning one scalar value.
+    lambda functions are advised against since the cache will not work with lambda
+    functions. sklearn_numba_dpex.common._utils expose some pre-defined
+    `fused_elementwise_func`s.
 
     Notes
     -----
@@ -283,8 +286,8 @@ def make_sum_reduction_2d_kernel(
     # axis, which motivates the need for different kernels with different work group
     # sizes for each cases. As a consequence, the shape of the intermediate results
     # in the main driver loop, and the `global_size` (total number of work items fired
-    # per call) for each kernel call, are also differents, and are variabilized with
-    # the lambda functions `get_result_shape1 and `get_global_size`.
+    # per call) for each kernel call, are also different, and are variabilized with
+    # the lambda functions `get_result_shape` and `get_global_size`.
     if axis == 1:
         work_group_size, kernels = _prepare_sum_reduction_2d_axis1(
             size0, work_group_size, fused_elementwise_func, dtype, device
@@ -311,6 +314,10 @@ def make_sum_reduction_2d_kernel(
             * work_group_size
             * math.ceil(size1 / sub_group_size)
         )
+
+    # The kernels seem to work fine with work_group_size==1 on GPU but fail on CPU.
+    if work_group_size == 1:
+        raise ValueError("work_group_size==1 is not supported.")
 
     # `fused_elementwise_func` is applied elementwise during the first pass on
     # data, in the first kernel execution only, using `fused_func_kernel`. Subsequent
@@ -385,7 +392,7 @@ def _prepare_sum_reduction_2d_axis1(
         check_power_of_2(work_group_size)
     else:
         # Round to the maximum smaller power of two
-        work_group_size = 2 ** (math.floor(math.log2(work_group_size)))
+        work_group_size = get_maximum_power_of_2_smaller_than(work_group_size)
 
     (
         reduction_block_size,
@@ -402,7 +409,7 @@ def _prepare_sum_reduction_2d_axis1(
         )
 
     return work_group_size, (
-        (partial_sum_reduction, partial_sum_reduction_nofunc),
+        (partial_sum_reduction, partial_sum_reduction_nofunc),  # kernels
         reduction_block_size,
     )
 
@@ -420,12 +427,12 @@ def _make_partial_sum_reduction_2d_axis1_kernel(
     two_as_a_long = np.int64(2)
 
     # Number of iteration in each execution of the kernel:
-    n_local_iterations = np.int64(math.floor(math.log2(work_group_size)) - 1)
+    n_local_iterations = np.int64(math.log2(work_group_size) - 1)
 
     local_values_size = work_group_size
     reduction_block_size = 2 * work_group_size
 
-    # ???: how does this strategy compares to having each thread reducing N contiguous
+    # ???: how does this strategy compare to having each thread reduce N contiguous
     # items ?
     @dpex.kernel
     # fmt: off
@@ -526,8 +533,8 @@ def _prepare_sum_reduction_2d_axis0(
     else:
         # Round work_group_size to the maximum smaller power-of-two multiple of
         # `sub_group_size`
-        n_sub_groups_per_work_group = 2 ** (
-            math.floor(math.log2(work_group_size / sub_group_size))
+        n_sub_groups_per_work_group = get_maximum_power_of_2_smaller_than(
+            work_group_size / sub_group_size
         )
         work_group_size = n_sub_groups_per_work_group * sub_group_size
 
@@ -546,7 +553,7 @@ def _prepare_sum_reduction_2d_axis0(
         )
 
     return work_group_size, (
-        (partial_sum_reduction, partial_sum_reduction_nofunc),
+        (partial_sum_reduction, partial_sum_reduction_nofunc),  # kernels
         reduction_block_size,
     )
 
@@ -564,9 +571,7 @@ def _make_partial_sum_reduction_2d_axis0_kernel(
     n_sub_groups_per_work_group = work_group_size // sub_group_size
 
     # Number of iteration in each execution of the kernel:
-    n_local_iterations = np.int64(
-        math.floor(math.log2(n_sub_groups_per_work_group)) - 1
-    )
+    n_local_iterations = np.int64(math.log2(n_sub_groups_per_work_group) - 1)
 
     local_values_size = (n_sub_groups_per_work_group, sub_group_size)
     reduction_block_size = 2 * n_sub_groups_per_work_group
@@ -668,10 +673,11 @@ def make_argmin_reduction_1d_kernel(size, device, dtype, work_group_size="max"):
         check_power_of_2(work_group_size)
     else:
         # Round to the maximum smaller power of two
-        work_group_size = 2 ** (math.floor(math.log2(work_group_size)))
+
+        work_group_size = get_maximum_power_of_2_smaller_than(work_group_size)
 
     # Number of iteration in each execution of the kernel:
-    n_local_iterations = np.int64(math.floor(math.log2(work_group_size)) - 1)
+    n_local_iterations = np.int64(math.log2(work_group_size) - 1)
 
     # TODO: the first call of partial_argmin_reduction in the final loop should be
     # written with only two arguments since "previous_result" does not exist yet.
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -441,17 +441,64 @@ def _make_partial_sum_reduction_2d_axis1_kernel(
         result,      # OUT       (n_rows, math.ceil(size / (2 * work_group_size),)
     ):
         # fmt: on
-        # NB: This kernel only perform a partial reduction
-        group_id = dpex.get_group_id(zero_idx)
-        local_work_id = dpex.get_local_id(zero_idx)
+        # NB: This kernel only perform a partial reduction.
 
-        sum_axis_size = summands.shape[minus_one_idx]
+        # The work groups are mapped to window of items in the input `summands` of size
+        # `(1, reduction_block_size)`, with
+        # `reduction_block_size = 2 * work_group_size`, such that the windows never
+        # overlap and create a grid that cover all the items. The work groups are
+        # indexed following a row-major order.
+
+        # `n_work_group_per_row` windows are necessary to cover one row of the input
+        # array.
         n_work_group_per_row = result.shape[minus_one_idx]
 
-        local_work_group_id_in_row = group_id % n_work_group_per_row
+        # The group of work items `group_id` is responsible for summing all items in
+        # the window it spans.
+        group_id = dpex.get_group_id(zero_idx)
+
+        # The work groups are indexed in row-major order, from that let's deduce the
+        # row of `summands` the window of items belongs to...
         row_idx = group_id // n_work_group_per_row
 
+        # ... and the position of the window within this row, ranging from 0 (first
+        # window in the row) to `n_work_group_per_row - 1` (last window in the row):
+        local_work_group_id_in_row = group_id % n_work_group_per_row
+
+        # Since all windows have size `reduction_block_size`, the position of the first
+        # item in the window is given by:
         first_value_idx = local_work_group_id_in_row * reduction_block_size
+
+        # To sum up, this work group `group_id` will sum items with coordinates
+        # `row_idx`, `col_idx`, with `col_idx` ranging from
+        # `first_value_idx` (first item in the window) to
+        # `first_value_idx + work_group_size - 1` (last item in the window)
+
+        # The current work item is indexed locally within the group of work items, with
+        # index `local_work_id` that can range from `0` (first item in the work group)
+        # to `work_group_size - 1` (last item in the work group)
+        local_work_id = dpex.get_local_id(zero_idx)
+
+        # Let's remember the size of the array to ensure that the last window in the
+        # row do not try to access items outside the buffer.
+        sum_axis_size = summands.shape[minus_one_idx]
+
+        # To begin with, each work item sums two items from the window that is spanned
+        # by this work group. The two items are chosen such that contiguous work items
+        # in the work group read contiguous items in the window, i.e the current work
+        # item reads a value at position `local_work_id` (relatively to the first item
+        # of the window), and the second value `work_group_size` items further
+        # (remember that `reduction_block_size = 2 * work_group_size`, so there are
+        # exactly two values to read for each work item).
+        # This memory access pattern is more efficient because:
+        # - it prevents bank conflicts (i.e serializing of write operations when
+        # threads access simultaneously close memory addresses in the same bank)
+        # - each sub group can coalesce the read operation, thus improving IO.
+        # Then this sum is written into a reserved slot in an array `local_values` in
+        # local memory of size `work_group_size` (i.e one slot for each work group and
+        # two items in the window), such that contiguous work items write into
+        # contiguous slots (yet again, this is a nicer memory access pattern).
+
         # NB: we use augend/addend vocabulary
         # in sum x + y, x is augend, y is addend
         # https://www.quora.com/What-is-Augend-and-Addend
@@ -460,10 +507,7 @@ def _make_partial_sum_reduction_2d_axis1_kernel(
 
         local_values = dpex.local.array(local_values_size, dtype=dtype)
 
-        # Each work item reads two value in global memory and sum it into the local
-        # memory
-        # NB: to optimize IO, consecutive work items within as same subgroup read
-        # consecutive adresses in memory.
+        # We must be careful to not read items outside of the array !
         if augend_idx >= sum_axis_size:
             local_values[local_work_id] = zero
         elif addend_idx >= sum_axis_size:
@@ -478,15 +522,24 @@ def _make_partial_sum_reduction_2d_axis1_kernel(
 
         dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
 
+        # Then, the sums of two scalars that have been written in `local_array` are
+        # further summed together into `local_array[0]`. At each iteration, half
+        # of the remaining work items are discarded and will be left idle, and the
+        # other half will sum together two value in the `local_array`, while using
+        # a similar memory access pattern than seen at the previous step.
         n_active_work_items = work_group_size
         for i in range(n_local_iterations):
-            # We discard half of the remaining active work items at each iteration
+            # At each iteration, half of the remaining work items with the highest id
+            # are discarded.
             n_active_work_items = n_active_work_items // two_as_a_long
             work_item_idx = first_value_idx + local_work_id + n_active_work_items
             if (
                 (local_work_id < n_active_work_items) and
                 (work_item_idx < sum_axis_size)
             ):
+                # Yet again, the remaining work items choose two values to sum such
+                # that contiguous work items read and write into contiguous slots of
+                # `local_values`.
                 local_values[local_work_id] += (
                     local_values[local_work_id + n_active_work_items]
                 )
@@ -584,34 +637,77 @@ def _make_partial_sum_reduction_2d_axis0_kernel(
         summands,    # IN        (sum_axis_size, n_cols)
         result,      # OUT       (math.ceil(size / (2 * reduction_block_size), n_cols)
     ):
-
         # fmt: on
         # NB: This kernel only perform a partial reduction
-        group_id = dpex.get_group_id(zero_idx)
-        local_work_id = dpex.get_local_id(zero_idx)
 
-        sum_axis_size = summands.shape[zero_idx]
+        # The work groups are mapped to window of items in the input `summands` of size
+        # `(reduction_block_size, sub_group_size)`,
+        # where `reduction_block_size = 2 * n_sub_groups_per_work_group` such that the
+        # windows never overlap and create a grid that cover all the items. The work
+        # groups are indexed following a column-major order.
+
+        # `n_blocks_per_col` windows are necessary to cover one column of the input
+        # array
         n_blocks_per_col = result.shape[zero_idx]
 
+        # The group of work items `group_id` is responsible for summing all items in
+        # the window it spans over axis 0, resulting in a output of shape
+        # `(1, sub_group_size)`
+        group_id = dpex.get_group_id(zero_idx)
+
+        # The work groups are indexed in column-major order. From this let's deduce the
+        # position of the window within the column...
         local_block_id_in_col = group_id % n_blocks_per_col
+
+        # Let's map the current work item to an index in a 2D grid, where the
+        # `work_group_size` work items are mapped in row-major order to the array
+        # of size `(n_sub_groups_per_work_group, sub_group_size)`.
+        local_work_id = dpex.get_local_id(zero_idx)      # 1D idx
+        local_row_idx = local_work_id // sub_group_size  # 2D idx, first coordinate
+        local_col_idx = local_work_id % sub_group_size   # 2D idx, second coordinate
+
+        # This way, each row in the 2D index can be seen as mapped to two rows in the
+        # corresponding window of items of the input `summands`, with the first row of
+        # work items being mapped to the first row of the input...
         first_row_idx = local_block_id_in_col * reduction_block_size
-        local_col_idx = local_work_id % sub_group_size
-        local_row_idx = local_work_id // sub_group_size
-        col_idx = (
-            (group_id // n_blocks_per_col) * sub_group_size + local_col_idx
-            )
+
+        # ... and the current work item, with index `local_row_idx`, being mapped to
+        # the following row of the input - which is also the row coordinate of the
+        # first term this work item is responsible to sum in the coming sum step:
+
         # NB: we use augend/addend vocabulary
         # in sum x + y, x is augend, y is addend
         # https://www.quora.com/What-is-Augend-and-Addend
         augend_row_idx = first_row_idx + local_row_idx
+
+        # The addend can be chosen arbitrarily (providing that different work items
+        # don't sum overlapping terms!). Let's use a mapping that is consistent with
+        # the choice in the kernel for axis 1, so that in both kernels the code reads
+        # familiar.
         addend_row_idx = augend_row_idx + n_sub_groups_per_work_group
 
+        # NB: because of how the 2D index on work items was defined, contiguous work
+        # items belonging to the same sub groups in the grid always span items in the
+        # data that are located in the same row, and so are contiguous, which ensures
+        # optimal memory access patterns by sub groups (refer to explanations in the
+        # kernel for axis 1 to understand how this access pattern is better). It is
+        # enough to ensure this pattern at the level of sub groups, and one can see
+        # that, contrarily to what is seen in the kernel for axis 1, it does not
+        # depends on how the pair (augend, addend) are chosen.
+
+        # The sums done by each work item during this first step will be written in
+        # this local array of size (n_sub_groups_per_work_group, sub_group_size)
         local_values = dpex.local.array(local_values_size, dtype=dtype)
 
-        # Each work item reads two value in global memory and sum it into the local
-        # memory
-        # NB: to optimize IO, consecutive work items within as same subgroup read
-        # consecutive adresses in memory.
+        # The current work item use the following second coordinate (given by the
+        # position of the window in the grid of windows, and by the local position of
+        # the work item in the 2D index):
+        col_idx = (
+            (group_id // n_blocks_per_col) * sub_group_size + local_col_idx
+            )
+
+        # We must be careful to not read items outside of the array !
+        sum_axis_size = summands.shape[zero_idx]
         if (col_idx >= n_cols) or (augend_row_idx >= sum_axis_size):
             local_values[local_row_idx, local_col_idx] = zero
         elif addend_row_idx >= sum_axis_size:
@@ -626,9 +722,15 @@ def _make_partial_sum_reduction_2d_axis0_kernel(
 
         dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
 
+        # Then, the sums of two scalars that have been written in `local_array` are
+        # further summed together into `local_array[0, :]`. At each iteration, half
+        # of the remaining work items are discarded and will be left idle, and the
+        # other half will sum together two value in the `local_array`, while using
+        # a similar memory access pattern than seen at the previous step.
         n_active_sub_groups = n_sub_groups_per_work_group
         for i in range(n_local_iterations):
-            # We discard half of the remaining active subgroups at each iteration
+            # At each iteration, half of the remaining work items with the highest id
+            # are discarded.
             n_active_sub_groups = n_active_sub_groups // two_as_a_long
             work_item_row_idx = first_row_idx + local_row_idx + n_active_sub_groups
             if (
@@ -642,9 +744,9 @@ def _make_partial_sum_reduction_2d_axis0_kernel(
 
             dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
 
-        # At this point local_values[0] + local_values[1] is equal to the sum of all
-        # elements in summands that have been covered by the work group, we write it
-        # into global memory
+        # At this point local_values[0, :] + local_values[1, :] is equal to the sum of
+        # all elements in summands that have been covered by the work group, we write
+        # it into global memory
         if (local_row_idx == zero_idx) and (col_idx < n_cols):
             result[local_block_id_in_col, col_idx] = (
                 local_values[zero_idx, local_col_idx] +
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -470,9 +470,9 @@ def _make_partial_sum_reduction_2d_axis1_kernel(
         first_value_idx = local_work_group_id_in_row * reduction_block_size
 
         # To sum up, this work group `group_id` will sum items with coordinates
-        # `row_idx`, `col_idx`, with `col_idx` ranging from
-        # `first_value_idx` (first item in the window) to
-        # `first_value_idx + work_group_size - 1` (last item in the window)
+        # (`row_idx`, `col_idx`), with `col_idx` ranging from `first_value_idx`
+        # (first item in the window) to `first_value_idx + work_group_size - 1` (last
+        # item in the window)
 
         # The current work item is indexed locally within the group of work items, with
         # index `local_work_id` that can range from `0` (first item in the work group)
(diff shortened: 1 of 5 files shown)


jjerphan:
FYI: I am waiting for #79 to be merged before reviewing this PR.

fcharras:
Ready for review

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+        n_block_for_col = result.shape[zero_idx]
+
+        local_block_id_in_col = (group_id % n_block_for_col)
+        first_row_idx = local_block_id_in_col * reduction_block_size
+        local_col_idx = (local_work_id % sub_group_size)
+        local_row_idx = (local_work_id // sub_group_size)
+        col_idx = (
+            group_id // n_block_for_col) * sub_group_size + local_col_idx
+        augend_row_idx = first_row_idx + local_row_idx
+        addend_row_idx = augend_row_idx + n_sub_group_per_work_group
```
ogrisel, line 519: Could you please insert a comment about what `augend` and `addend` stand for? I was not familiar with this naming convention myself and I had to google it.

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -32,17 +32,28 @@
     ],
```
ogrisel, line 32: Let's parametrize the test by shape tuple only and do the `dpt.arange` call in the body of the test instead. The expected result can be computed by numpy instead of passing it as a test parameter and using the transposition trick.

This way it will be easier to run the test for more shapes, including more edge cases such as `(1, 1)`, `(1, 2)` and `(3, 1)` and maybe some larger shapes.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+
+        local_values = dpex.local.array(local_values_size, dtype=dtype)
+
+        # Each work item reads two value in global memory and sum it into the local
+        # memory
```
ogrisel, line 524: Could you please make it explicit how those two values are layout in global memory? I would have naively assumed they are in consecutive rows of the 2d input data array but by reading the code, I see they are separated by `n_sub_group_per_work_group` rows but I cannot grasp the motivation.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
 
     Notes
     -----
-    `work_group_size` is assumed to be a power of 2.
+    if `size1` is None then the kernel expects 1d tensor inputs, and `axis` parameter
+    is ignored.
```
jjerphan, line 239:
```suggestion
    If `size1` is None then the kernel expects 1d tensor inputs, and `axis` parameter
    is ignored.
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
 
-    if `size1` is None then the kernel expects 1d tensor inputs. If `size1` is not None
-    then the expected shape of input tensors is `(size0, size1)`, and the reduction
-    operation is equivalent to input.sum(axis=1). Depending on the size of the sum
-    axis, it might be worth tuning `work_group_size` for better performance.
+    If `size1` is not None then the expected shape of input tensors is
+    `(size0, size1)`, and the reduction operation is equivalent to
+    `input.sum(axis=axis)`.
+
+    if `size1` is None or if `axis` is `0`, then `work_group_size` is assumed to be a
```
jjerphan, line 245:
```suggestion
    If `size1` is None or if `axis` is `0`, then `work_group_size` is assumed to be a
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
+    power of 2, and the parameter `sub_group_size` is ignored.
+
+    if `size1` is not `None` and `axis` is `1`, then `work_group_size` is assumed to be
```
jjerphan, line 248:
```suggestion
    If `size1` is not `None` and `axis` is `1`, then `work_group_size` is assumed to be
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+                f"work_group_size={work_group_size}"
+            )
+        check_power_of_2(
+            n_sub_group_per_work_group := (work_group_size // sub_group_size)
+        )
+
+    else:
+        # Round work_group_size to the maximum smaller power-of-two multiple of
+        # `sub_group_size`
+        n_sub_group_per_work_group = 2 ** (
```
ogrisel, line 484: `n_sub_groups_per_work_group`?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+        result,      # OUT       (n_rows, math.ceil(size / (2 * reduction_block_size),)
+    ):
+
+        # fmt: on
+        # NB: This kernel only perform a partial reduction
+        group_id = dpex.get_group_id(zero_idx)
+        local_work_id = dpex.get_local_id(zero_idx)
+
+        sum_axis_size = summands.shape[zero_idx]
+        n_block_for_col = result.shape[zero_idx]
```
ogrisel, line 510: `n_blocks_per_col`?

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -32,17 +32,28 @@
...
+        array_in = dpt.asarray(array_in.T, order="C")
+
     array_in = dpt.astype(array_in, dtype)
 
-    device = dpctl.SyclDevice()
+    device = array_in.device.sycl_device
+
+    if work_group_size == "max":
+        sub_group_size = min(device.sub_group_sizes)
+
```
jjerphan, line 45:
```suggestion
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+        summands,    # IN        (n_rows, sum_axis_size)
+        result,      # OUT       (n_rows, math.ceil(size / (2 * reduction_block_size),)
+    ):
+
+        # fmt: on
+        # NB: This kernel only perform a partial reduction
+        group_id = dpex.get_group_id(zero_idx)
+        local_work_id = dpex.get_local_id(zero_idx)
+
+        sum_axis_size = summands.shape[zero_idx]
```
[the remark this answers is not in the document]
ogrisel: I think I would rather rename `sum_axis_size` to `shape_0` or `summands_shape_0` if you prefer to be extra explicit.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
+    a multiple of `sub_group_size`, such that `work_group_size // sub_group_size` is
+    a power of 2.
+
+    Depending on the size of the sum axis, it might be worth tuning `work_group_size`
+    for better performance. When `axis` is `1`, tuning `sub_group_size` with respect
+    to the size of the other axis might also be beneficial.
     """
-    if size1 is not None:
-        n_rows = size0
+    if is_1d := not size1:
```
jjerphan, line 256:
```suggestion
    if is_1d := not size1:
        # Makes the 1d case a special 2d case to reuse the same kernel.
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
+        work_group_size, kernels = _make_partial_sum_reduction_2d_axis1_kernels(
+            size0, work_group_size, fused_unary_func, dtype, device
+        )
         sum_axis_size = size1
+        get_result_shape = lambda result_sum_axis_size: (size0, result_sum_axis_size)
+        get_global_size = (
+            lambda result_sum_axis_size: result_sum_axis_size * work_group_size * size0
+        )
+
     else:
```
jjerphan, line 271:
```suggestion
    else:  # axis == 2, necessarily for now
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
+
         sum_axis_size = size0
+        get_result_shape = lambda result_sum_axis_size: (result_sum_axis_size, size1)
+        get_global_size = (
+            lambda result_sum_axis_size: result_sum_axis_size
+            * work_group_size
+            * math.ceil(size1 / sub_group_size)
+        )
 
     # fused_unary_func is applied elementwise during the first pass on data, in the
```
jjerphan, line 285:
```suggestion
    # `fused_unary_func` is applied elementwise during the first pass on data, in the
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
+        size0 = 1
+
+    if axis == 1:
+        work_group_size, kernels = _make_partial_sum_reduction_2d_axis1_kernels(
+            size0, work_group_size, fused_unary_func, dtype, device
+        )
         sum_axis_size = size1
+        get_result_shape = lambda result_sum_axis_size: (size0, result_sum_axis_size)
+        get_global_size = (
+            lambda result_sum_axis_size: result_sum_axis_size * work_group_size * size0
```
jjerphan, line 268: Why do we not use this?
```suggestion
            lambda result_sum_axis_size: result_sum_axis_size 
            * work_group_size 
            * math.ceil(size0 / sub_group_size)
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+                summands[augend_row_idx, col_idx])
+        else:
+            local_values[local_row_idx, local_col_idx] = fused_unary_func_(
+                summands[augend_row_idx, col_idx]) + fused_unary_func_(
+                    summands[addend_row_idx, col_idx])
+
+        dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+
+        current_n_nows = n_sub_group_per_work_group
+        for i in range(local_n_iterations):
```
ogrisel, line 538: I would have found it more natural to name this `n_local_iterations`.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
                 local_values[zero_idx] + local_values[one_idx]
             )
 
-    return work_group_size, partial_sum_reduction
+    if return_only_kernel:
+        return partial_sum_reduction
+
+    partial_sum_reductiom_nofunc = _make_partial_sum_reduction_2d_axis1_kernels(
+        n_rows, work_group_size, None, dtype, device, return_only_kernel=True
+    )
```
jjerphan, line 437: This recursion looks unintuitive to me.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+
+    partial_sum_reductiom_nofunc = _make_partial_sum_reduction_2d_axis0_kernels(
+        n_cols,
+        work_group_size,
+        sub_group_size,
+        None,
+        dtype,
+        device,
+        return_only_kernel=True,
+    )
```
jjerphan, line 569: Same comment regarding this recursion.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+
+        dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+
+        current_n_nows = n_sub_group_per_work_group
+        for i in range(local_n_iterations):
+            # We discard half of the remaining active subgroups at each iteration
+            current_n_nows = current_n_nows // two_as_a_long
+            if (local_row_idx < current_n_nows) and (
+                    col_idx < n_cols) and (
+                        first_row_idx + local_row_idx + current_n_nows) < sum_axis_size:
```
ogrisel, line 543: Is this really black formatted? I would have expected something like the following or similar:

```suggestion
            if (
                (local_row_idx < current_n_nows) and
                (col_idx < n_cols) and
                (first_row_idx + local_row_idx + current_n_nows) < sum_axis_size
            ):
                
```

jjerphan (commented):
A first review. The recursive kernel factory looks hard to understand to me at first sight.

sklearn_numba_dpex/common/kernels.py
ogrisel: not sure about the position of the `and`s though.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+
+            dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+
+        # At this point local_values[0] + local_values[1] is equal to the sum of all
+        # elements in summands that have been covered by the work group, we write it
+        # into global memory
+        if (local_row_idx == zero_idx) and (col_idx < n_cols):
+            result[local_block_id_in_col, col_idx] = (
+                local_values[zero_idx, local_col_idx] + local_values[
+                    one_idx, local_col_idx])
```
ogrisel, line 556: again the formatting does not look blackish to me. I would have expected something more readable like:

```suggestion
            result[local_block_id_in_col, col_idx] = (
                local_values[zero_idx, local_col_idx]
                + local_values[one_idx, local_col_idx]
            )
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+    # ???: how does this strategy compares to having each thread reducing N contiguous
+    # items ?
+    @dpex.kernel
+    # fmt: off
+    def partial_sum_reduction(
+        summands,    # IN        (n_rows, sum_axis_size)
+        result,      # OUT       (n_rows, math.ceil(size / (2 * reduction_block_size),)
+    ):
+
+        # fmt: on
```
ogrisel, line 504: I suspect that the `fmt: on` marker does not work as expected because many multiline expressions looks unreadable to my black-trained eyes.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+
+        sum_axis_size = summands.shape[zero_idx]
+        n_block_for_col = result.shape[zero_idx]
+
+        local_block_id_in_col = (group_id % n_block_for_col)
+        first_row_idx = local_block_id_in_col * reduction_block_size
+        local_col_idx = (local_work_id % sub_group_size)
+        local_row_idx = (local_work_id // sub_group_size)
+        col_idx = (
+            group_id // n_block_for_col) * sub_group_size + local_col_idx
```
ogrisel, line 517: again this does not look blackish.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+            local_values[local_row_idx, local_col_idx] = fused_unary_func_(
+                summands[augend_row_idx, col_idx])
+        else:
+            local_values[local_row_idx, local_col_idx] = fused_unary_func_(
+                summands[augend_row_idx, col_idx]) + fused_unary_func_(
+                    summands[addend_row_idx, col_idx])
+
+        dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
+
+        current_n_nows = n_sub_group_per_work_group
```
ogrisel, line 537: `current_n_rows`? or `n_active_local_rows`?

ogrisel (commented):
Here is a pass of review. I did not find any fundamental problem but I have had a hard time understanding the code because of a mix of formatting problems, variable names that could be improved and maybe lack of big picture inline comment at the beginning of the kernel code.

sklearn_numba_dpex/common/kernels.py
ogrisel: `axis == 0`

sklearn_numba_dpex/common/kernels.py
fcharras: When axis=1, each work group perform a local reduction on axis 1 in a window of size (1, 2 * work_group_size). (does not depend on sub_group_size)

When axis=0, each work group perform a local reduction on axis 0 in a window of size (sub_group_size_,work_group_size // sub_group_size).

That's why the global size and the size of the intermediate results are different in both cases.

And the reason for those differences is that it's done to optimize the IO which I think is the main bottleneck, and the IO pattern is different in both cases, so are the kernels.

I added some more comments explaining that.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -228,31 +235,57 @@ def make_sum_reduction_2d_axis1_kernel(
...
+    if `size1` is None or if `axis` is `0`, then `work_group_size` is assumed to be a
+    power of 2, and the parameter `sub_group_size` is ignored.
+
+    if `size1` is not `None` and `axis` is `1`, then `work_group_size` is assumed to be
+    a multiple of `sub_group_size`, such that `work_group_size // sub_group_size` is
+    a power of 2.
+
+    Depending on the size of the sum axis, it might be worth tuning `work_group_size`
+    for better performance. When `axis` is `1`, tuning `sub_group_size` with respect
+    to the size of the other axis might also be beneficial.
```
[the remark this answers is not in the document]
fcharras: `fused_unary_func` is already documented in one of the blocks above `Notes`. OK for renaming.

```
    If `fused_elementwise_func` is not None, it will be applied element-wise before 
    summing. It must be a function that will be interpreted as a dpex.func and is
    subject to the same rules. It is expected to take one scalar argument and returning
    one scalar value. lambda functions are advised against since the cache will not 
    work with lambda functions. sklearn_numba_dpex.common._utils expose some 
    pre-defined `fused_elementwise_func`s.
```

Adding memory layouts expectations:

>     This implementation is optimized for C-contiguous arrays.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+@lru_cache
+def _make_partial_sum_reduction_2d_axis0_kernels(
+    n_cols,
+    work_group_size,
+    sub_group_size,
+    fused_unary_func,
+    dtype,
+    device,
+    return_only_kernel=False,
+):
```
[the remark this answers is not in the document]
fcharras: There's already a docstring that sums up the approach in the top level function `make_sum_reduction_2d_kernel`:
```
    The strategy consists in performing local reductions in each work group using local
    memory where each work item combine two values, thus halving the number of values,
    and the number of active work items. At each iteration the work items are discarded
    in a bracket manner. The work items with the greatest ids are discarded first, and
    we rely on the fact that the remaining work items are adjacents to optimize the RW
    operations.

    Once the reduction is done in a work group the result is written in global memory,
    thus creating an intermediary result whose size is divided by
    `2 * work_group_size` if axis=1, or by `2 * work_group_size // sub_group_size` if\
    axis=0. This is repeated as many times as needed until only one value remains in
    global memory.
```

I think it's better there since it applies to implementations for both axis ?

For more in-depth explanations, since there are many technical details, I propose instead citing litterature. I haven't found a piece that really stands out as a better read than the others, but at least it converges on the same ideas:

```
    The algorithmic approach that is chosen to implement the underlying kernels takes
    inspiration from many sources that are available online, such as [1]_, [2]_, [3]_ .
    
    .. [1] Timcheck, S. W. (2017). Efficient Implementation of Reductions on GPU 
    Architectures.
    
    .. [2] https://dournac.org/info/gpu_sum_reduction
    
    .. [3] https://shreeraman-ak.medium.com/parallel-reduction-with-cuda-d0ae10c1ae2c
```

sklearn_numba_dpex/common/kernels.py
fcharras: typo :facepalm: was meant to be `current_n_rows`, but let's rename it `n_active_subgroups`.

Also applied similar renaming for `axis1` sum.

sklearn_numba_dpex/common/kernels.py
fcharras: Although you're right your suggestion is prettier (and in fact @jjerphan suggested the same in the PR for the cleaning of the `axis1` sum, and I forgot to apply it for `axis0` again :sob:  ), the former _is_ black formatted, I don't think there's an issue with `fmt` tags here. Answer applies to other similar comments :grin:

sklearn_numba_dpex/common/kernels.py
```diff
@@ -198,10 +198,17 @@ def half_l2_norm(
 # TODO: this kernel could be abstracted away to support other commutative binary
 # operators than sum.
 @lru_cache
-def make_sum_reduction_2d_axis1_kernel(
-    size0, size1, device, dtype, work_group_size="max", fused_unary_func=None
+def make_sum_reduction_2d_kernel(
+    size0,
+    size1,
```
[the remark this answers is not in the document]
fcharras: I'm neutral myself. Up to you if you want it pushed in this PR. Else i'd let it like it is.

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
fcharras: You're right the comment was wrong. It is fixed:

```        summands,    # IN        (sum_axis_size, n_cols)
        result,      # OUT       (math.ceil(size / (2 * reduction_block_size), n_cols)
```

Regarding naming for dimensions and sizes my idea was to use `size0`, `size1` as long as the number of dimension of the input is unknown (we can have size1 == None), and when it's known use `n_row`, `n_col` since it's 2d.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+        group_id = dpex.get_group_id(zero_idx)
+        local_work_id = dpex.get_local_id(zero_idx)
+
+        sum_axis_size = summands.shape[zero_idx]
+        n_block_for_col = result.shape[zero_idx]
+
+        local_block_id_in_col = (group_id % n_block_for_col)
+        first_row_idx = local_block_id_in_col * reduction_block_size
+        local_col_idx = (local_work_id % sub_group_size)
+        local_row_idx = (local_work_id // sub_group_size)
```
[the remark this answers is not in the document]
fcharras: Variables prefixed by `local_` refers to position in the array of allocated local memory, and variables without the prefix refers to positions in global memory.

`row` refers to first coordinate, and `col` to second coordinate.

sklearn_numba_dpex/common/kernels.py
fcharras: The memory accesses are done so that contiguous work items in a given subgroup access contiguous memory addresses. In theory that should optimize IO since RW access with this patterns are coalesced.

The different sources I've added also converge on using this pattern, and, although we've never verified its impact, we rely on it for IO in all our kernels, so I don't think it's particularly needed to insist on it here. The top level docstring hints at an explanation already:

```
    we rely on the fact that the remaining work items are adjacents to optimize the RW
    operations.
```

fcharras:
TY for the reviews @jjerphan @ogrisel I've applied your suggestions and answered the questions for everything except unit tests, which will come in a future commit.

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -1,16 +1,16 @@
...
 from sklearn_numba_dpex.common.kernels import (
     make_argmin_reduction_1d_kernel,
-    make_sum_reduction_2d_axis1_kernel,
+    make_sum_reduction_2d_kernel,
 )
 from sklearn_numba_dpex.testing.config import float_dtype_params
 
 
+@pytest.mark.parametrize("axis", [0, 1])
 @pytest.mark.parametrize("work_group_size", [2, 4, 8, "max"])
```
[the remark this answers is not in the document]
fcharras: The tests work with wgs=1 so apparently yes we can keep the test.

sklearn_numba_dpex/common/tests/test_kernels.py
fcharras: Adding edge cases.

The transpose trick is nice because we only have to define the test cases for one axis and copy for the other axis no ?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +429,149 @@ def partial_sum_reduction(
...
+        n_sub_group_per_work_group = 2 ** (
+            math.floor(math.log2(work_group_size / sub_group_size))
+        )
+        work_group_size = n_sub_group_per_work_group * sub_group_size
+
+    # Number of iteration in each execution of the kernel:
+    local_n_iterations = np.int64(math.floor(math.log2(n_sub_group_per_work_group)) - 1)
+
+    local_values_size = (n_sub_group_per_work_group, sub_group_size)
+    reduction_block_size = 2 * n_sub_group_per_work_group
```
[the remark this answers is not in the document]
fcharras: In the end it's only a sum, were one of those things wrong I think it would be quick to show up in the unit tests that check outputs values (and I've experienced it already). Honestly I think it's complicated and unnecessary.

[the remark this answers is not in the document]
ogrisel: My point is that this code implements some logic to choose good sizes from a performance point of view and this code is not trivial and it would be good to check that the expected sizing logic is the one intended.

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
fcharras: In fact it passes on gpu but not on cpu. I don't think it's worth debugging that :thinking:

sklearn_numba_dpex/common/tests/test_kernels.py
ogrisel: I find it not nice because it's harder to reason about what is the intention of the test and what is contiguous in memory.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -336,8 +387,40 @@ def fused_unary_func_(x):
...
+        partial_sum_reduction_nofunc = partial_sum_reduction
+    else:
+        _, partial_sum_reduction_nofunc = _make_partial_sum_reduction_2d_axis1_kernel(
+            n_rows, work_group_size, fused_elementwise_func_, dtype
+        )
+
+    return work_group_size, (
+        (partial_sum_reduction, partial_sum_reduction_nofunc),
+        reduction_block_size,
+    )
```
jjerphan, line 407:
```suggestion
    return (
        work_group_size,
        (partial_sum_reduction, partial_sum_reduction_nofunc),  # kernels
        reduction_block_size,
    )
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -406,7 +494,159 @@ def partial_sum_reduction(
...
+        partial_sum_reduction_nofunc = partial_sum_reduction
+    else:
+        _, partial_sum_reduction_nofunc = _make_partial_sum_reduction_2d_axis0_kernel(
+            n_cols, work_group_size, sub_group_size, fused_elementwise_func_, dtype
+        )
+
+    return work_group_size, (
+        (partial_sum_reduction, partial_sum_reduction_nofunc),
+        reduction_block_size,
+    )
```
jjerphan, line 551:
```suggestion
    return (
        work_group_size,
        (partial_sum_reduction, partial_sum_reduction_nofunc),  # kernels
        reduction_block_size,
    )
```

fcharras:
Added tests for edge cases, and for invalid work group or sub group sizes.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -223,43 +232,90 @@ def make_sum_reduction_2d_axis1_kernel(
...
-    # first kernel execution only.
-    work_group_size, fused_func_kernel = _make_partial_sum_reduction_2d_axis1_kernel(
-        n_rows, work_group_size, fused_unary_func, dtype, device
-    )
-    # subsequent kernel calls only sum the data.
-    _, nofunc_kernel = _make_partial_sum_reduction_2d_axis1_kernel(
-        n_rows, work_group_size, fused_unary_func=None, dtype=dtype, device=device
-    )
+    else:  # axis == 0
+        work_group_size, kernels = _prepare_sum_reduction_2d_axis0(
```
jjerphan, line 299: I find this unpacking unconventional and here my proposal to improve it (2/3).
```suggestion
        (
            work_group_size,
            kernels, 
            reduction_block_size,
        ) = _prepare_sum_reduction_2d_axis0(
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -223,43 +232,90 @@ def make_sum_reduction_2d_axis1_kernel(
...
+            lambda result_sum_axis_size: result_sum_axis_size
+            * work_group_size
+            * math.ceil(size1 / sub_group_size)
+        )
 
-    check_power_of_2(work_group_size)
+    # `fused_elementwise_func` is applied elementwise during the first pass on
+    # data, in the first kernel execution only, using `fused_func_kernel`. Subsequent
+    # kernel calls only sum the data, using `nofunc_kernel`.
+    (fused_func_kernel, nofunc_kernel), reduction_block_size = kernels
```
jjerphan, line 318: I find this unpacking unconventional and here my proposal to improve it (3/3).
```suggestion
    (fused_func_kernel, nofunc_kernel)  = kernels
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -223,43 +232,90 @@ def make_sum_reduction_2d_axis1_kernel(
...
+    # efficiently requires to adapt to very different IO patterns depending on the sum
+    # axis, which motivates the need for different kernels with different work group
+    # sizes for each cases. As a consequence, the shape of the intermediate results
+    # in the main driver loop, and the `global_size` (total number of work items fired
+    # per call) for each kernel call, are also differents, and are variabilized with
+    # the lambda functions `get_result_shape1 and `get_global_size`.
+    if axis == 1:
+        work_group_size, kernels = _prepare_sum_reduction_2d_axis1(
+            size0, work_group_size, fused_elementwise_func, dtype, device
+        )
```
jjerphan, line 291: I find this unpacking unconventional and here my proposal to improve it (1/3).
```suggestion
        (
            work_group_size,
            kernels, 
            reduction_block_size,
        ) = _prepare_sum_reduction_2d_axis1(
```

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -6,11 +8,12 @@
 
 from sklearn_numba_dpex.common.kernels import (
     make_argmin_reduction_1d_kernel,
-    make_sum_reduction_2d_axis1_kernel,
+    make_sum_reduction_2d_kernel,
 )
 from sklearn_numba_dpex.testing.config import float_dtype_params
 
 
+@pytest.mark.parametrize("axis", [0, 1])
```
jjerphan, line 16: I guess `axis` and `expected_result` are coupled and better be joined in the same parametrization mark, right?

jjerphan (commented):
This is my last review.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -216,43 +225,90 @@ def make_sum_reduction_2d_axis1_kernel(
...
 
-    If fused_unary_func is not None, it will be applied element-wise before summing.
-    It must be a function that will be interpreted as a dpex.func and is subject to the
-    same rules. It is expected to take one scalar argument and returning one scalar
-    value. lambda functions are advised against since the cache will not work with
-    lambda functions. sklearn_numba_dpex.common._utils expose some pre-defined
-    `fused_unary_funcs`.
+    If `fused_elementwise_func` is not None, it will be applied element-wise before
+    summing. It must be a function that will be interpreted as a dpex.func and is
+    subject to the same rules. It is expected to take one scalar argument and returning
```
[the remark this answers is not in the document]
fcharras: let's just paste https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html then

sklearn_numba_dpex/common/kernels.py
```diff
@@ -397,7 +485,157 @@ def partial_sum_reduction(
...
+    # Number of iteration in each execution of the kernel:
+    n_local_iterations = np.int64(
+        math.floor(math.log2(n_sub_groups_per_work_group)) - 1
+    )
+
+    local_values_size = (n_sub_groups_per_work_group, sub_group_size)
+    reduction_block_size = 2 * n_sub_groups_per_work_group
+
+    # ???: how does this strategy compares to having each thread reducing N contiguous
+    # items ?
```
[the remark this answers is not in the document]
fcharras: same :+1:

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
ogrisel: My objective would be to make the reader of each individual kernel able to straightforwardly understand how the parallelism work without having to read any external reference.

It's good to have a high level description with reference in the top level API but it would be even better to also have short and concrete and specific inline comments in the code of the kernel to avoid any indirection when reading it.

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
ogrisel: At least let's add an inline comment to document the problem. If you can find a minimal reproducer that would be great to get it fixed upstream so that we don't encounter the same bug again 6 months later down the road.

sklearn_numba_dpex/common/kernels.py
ogrisel: > the former is black formatted, I don't think there's an issue with fmt tags here. Answer applies to other similar comments 😁

I checked and black is disabled in this block. I don't know how to make the `fmt on/off` things to work to disable black only inside the prototype of the function unfortunately.

ogrisel: I think I found a solution: removing the `# fmt: off` / `# fmt: on` lines and instead appending a `# fmt: skip` at the end of the line with the `):` of the function prototype seems to work. Let me try to update this file accordingly.

ogrisel: Hum, it works when using the command line manually but then the pre-commit command fails... Let me investigate.

ogrisel: The version of black in the pre-commit config is probably too old to handle this pattern. I will re-run experiments with both versions after lunch.

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
fcharras: > My objective would be to make the reader of each individual kernel able to straightforwardly understand how the parallelism work without having to read any external reference.

I disagree on this objective. I think readers that arrive at this (low) level of code and need to maintain, debug, or improve, or formulate criticism for some reason, should pass some minimal level of required knowledge on gpu programming. We're already way above average level of commenting per line of code in this repository. I see the point for difficult kernels such as lloyd or topk and I happily provide more commenting there, since even with resources it requires some mental gymnastics to mirror the code and the papers, but this kernel sum is much simpler. 

So maybe you have other readers in mind ? such as students ? but if these implementations are going to serve as educational content it should be explicited as a target of the project. I don't have this in mind, and I think it would slow down development overall.

Maybe we can open a separate issue to discuss this and start contributing standards for the repo ?

Another thing we can discuss is how much time of review is considered OK per line of code in a PR.

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
fcharras: Let me add it to my bucket list of leftover task that I clean in a single PR from time to time ?

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
fcharras: By the way @ogrisel if you want more perspective on the approach but don't have time to skim through the papers I'd happily provide you with live explanations, maybe we can just browse the code together.

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
jjerphan: Works for me, that's not urgent to me.

sklearn_numba_dpex/common/tests/test_kernels.py
jjerphan: Yes, but for some cases, reducing on another axis should given different results, right?

For instance, the first third parametrization case for `array_in, expected_result` must when `axis==0` be:

```
array([15.0, 18.0, 21.0, 24.0, 27.0])
```

right?

sklearn_numba_dpex/common/kernels.py
fcharras: Thanks for the investigation !!

fcharras:
Latest commit aims at addressing the remaining comments.

sklearn_numba_dpex/common/tests/test_kernels.py
fcharras: I rewrote the test as you said

sklearn_numba_dpex/common/tests/test_kernels.py
fcharras: No the test wasn't supposed to be read like this.

But since you and @ogrisel had the same remark I ended up rewriting it, hope that can solve the confusion about axis and expected values.

jjerphan (approved):
LGTM, let's move on.

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
fcharras: I've added exhaustive commenting so the PR can move on.

I'm :-1: for doing it again in the future for this amount of lines ~ took almost half a day and I think the reader still needs a pen and a paper to understand what is going on (in all the kernels the grid of work groups is mapped to the data is a way that a drawing makes it a lot more telling than words).

sklearn_numba_dpex/common/tests/test_kernels.py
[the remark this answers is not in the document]
ogrisel: Was this comment addressed?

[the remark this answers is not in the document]
fcharras: yes, kernel now raises an error for `work_group_size == 1` https://github.com/soda-inria/sklearn-numba-dpex/pull/80/files#diff-326118b7b149309370de9833d674e45b523445573056b66b14daed1793f00fe2R320

sklearn_numba_dpex/common/tests/test_kernels.py
```diff
@@ -6,43 +8,42 @@
...
 
-    sum_reduction_2d_kernel = make_sum_reduction_2d_axis1_kernel(
+    array_in = dpt.asarray(array_in, order="C")
+
+    device = array_in.device.sycl_device
+
+    if work_group_size == "max":
+        sub_group_size = min(device.sub_group_sizes)
+    elif work_group_size == 1:
+        sub_group_size = 1
```
ogrisel, line 35: This is never tested in the current parametrization.

fcharras: indeed the `elif` can be removed

fcharras:
Why does the test start failing only now ? I guess it would be related to https://github.com/soda-inria/sklearn-numba-dpex/blob/7e651390c29cf7917daa8314e0b1f667ef22656c/sklearn_numba_dpex/common/kernels.py#L319 that has been merged but this branch doesn't include the change yet ? :thinking:

jjerphan:
@fcharras, should we merge `main` in the branch of this PR?

fcharras:
Yes, if what's happening is that github runs the tests after merging on main (?), main needs to be merged first so the test can be adjusted.

fcharras:
I'll do it, but was waiting for @ogrisel to finish review.

ogrisel (approved):
Thanks for the more detailed comments and the improvements in the coverage and readability of the tests.

I would have more suggestions to try to improve the readability of the logic for new comers but let's merge this PR as it is and instead let me open follow-up PRs to the main branch

fcharras:
Thanks for review ! That would be really cool if you can open follow-up PRs. :+1:

PR #81, opened 2022-12-28
FEA 1D Radix `topk` and application to cluster relocation in `KMeans`

fcharras:


fcharras:
script actually working now !

fcharras:
Made the script more readable !

fcharras:
Out of WIP ! also, the top k is now used in place of `dpnp.partition` for kmeans cluster relocation step

jjerphan:
Must https://github.com/soda-inria/sklearn-numba-dpex/pull/79 and https://github.com/soda-inria/sklearn-numba-dpex/pull/80 be merged before this PR?

fcharras:
yes, but there isn't too much risk of conflicts when iterating on those. #79 and #80 implement sum on axis 0, which in the end is used only once in this pr.

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
```diff
@@ -212,104 +210,24 @@ def test_relocate_empty_clusters(dtype):
...
     assert_allclose(
-        kmeans_vanilla.cluster_centers_, asnumpy(kmeans_engine.cluster_centers_)
+        kmeans_vanilla.cluster_centers_[vanilla_labels],
+        asnumpy(kmeans_engine.cluster_centers_)[engine_labels],
     )
     assert_allclose(kmeans_vanilla.inertia_, kmeans_engine.inertia_)
 
 
-@pytest.mark.parametrize("dtype", float_dtype_params)
-def test_select_samples_far_from_centroid_kernel(dtype):
```
jjerphan, line 227: Why was this test removed?

sklearn_numba_dpex/common/tests/test_topk.py
```diff
@@ -0,0 +1,91 @@
...
+
+from sklearn_numba_dpex.common.topk import topk, topk_idx
+from sklearn_numba_dpex.testing import _assert_array_equal_any
+from sklearn_numba_dpex.testing.config import float_dtype_params
+
+
+@pytest.mark.parametrize("dtype", float_dtype_params)
+@pytest.mark.parametrize("work_group_size", [4, 8, None])
+@pytest.mark.parametrize(
+    "array_in, expected_top_3, list_of_expected_top_3_idx",
```
jjerphan, line 13: Is this more exact?
```suggestion
    "array_in, expected_top_3, list_of_possible_top_3_idx",
```

Also is there any specific tie breaking for this implementation? E.g. does it choose the smallest indices in case of tie break?

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
+# This implementation takes inspiration from other open-source implementations, such as
+# [1] https://github.com/pytorch/pytorch/blob/master/caffe2/operators/\
+#    top_k_radix_selection.cuh
```
jjerphan, line 3:
```suggestion
# [1] https://github.com/pytorch/pytorch/blob/master/caffe2/operators/top_k_radix_selection.cuh  # noqa
```

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+# or
+# [2] https://developer.nvidia.com/blog/gpu-pro-tip-fast-histograms-using-shared\
+# -atomics-maxwell
```
jjerphan, line 6:
```suggestion
# [2] https://developer.nvidia.com/blog/gpu-pro-tip-fast-histograms-using-shared-atomics-maxwell  # noqa
```

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+
+# TODO: apply Dr. TopK optimizations
```
jjerphan, line 8:
```suggestion
# TODO: apply Dr. TopK optimizations (https://arxiv.org/abs/2109.08219)
```

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+#      a float is used to encode an unsigned integer. Said differently, first transform
+#      the float to the integer such as the bits used to encode the integer are the
+#      same than the bits used to encode the float)
+#    - if the float item is positive, set the first bit of the unsigned integer to 1
+#    - if the float item is negative, flip all of the bits of the unsigned integer
+#      (i.e set to 1 if bit is 0, else 0)
+#
+# See also:
+# https://stackoverflow.com/questions/4640906/radix-sort-sorting-a-float-data/\
+#     4641059#4641059
```
jjerphan, line 236:
```suggestion
# https://stackoverflow.com/questions/4640906/radix-sort-sorting-a-float-data/4641059#4641059  # noqa
```

jjerphan, line 236: Should all of those rationales comes at the top of this file?

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+        array_in, k, group_sizes
+    )
+    result = dpt.empty(sh=(k,), dtype=np.int64, device=device)
+    result_idx = dpt.zeros(sh=(1,), dtype=np.int32)
+    gather_topk_kernel = _make_gather_topk_idx_kernel(
+        array_in.shape[0], k, work_group_size
+    )
+
+    gather_topk_kernel(array_in, threshold, k_in_subset, result_idx, result)
+    return result
```
jjerphan, line 57: Have I missed something?
```suggestion
    return result_idx
```

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+    k_in_subset, threshold, work_group_size, dtype, device = _get_topk_threshold(
+        array_in, k, group_sizes
+    )
+    result = dpt.empty(sh=(k,), dtype=np.int64, device=device)
+    result_idx = dpt.zeros(sh=(1,), dtype=np.int32)
+    gather_topk_kernel = _make_gather_topk_idx_kernel(
+        array_in.shape[0], k, work_group_size
+    )
+
+    gather_topk_kernel(array_in, threshold, k_in_subset, result_idx, result)
```
jjerphan, line 56:
```suggestion
    gather_topk_idx_kernel = _make_gather_topk_idx_kernel(
        array_in.shape[0], k, work_group_size
    )

    gather_topk_idx_kernel(array_in, threshold, k_in_subset, result_idx, result)
```

jjerphan (commented):
Thank you, @fcharras!

Here is a first pass on all but the core of `sklearn_numba_dpex/common/topk.py`.

Note that I will prefix the title of this PR for meeting scikit-learn's conventions.

sklearn_numba_dpex/common/topk.py
fcharras: i'll rename result_idx to index_buffer maybe ?

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+    make_initialize_to_zeros_2d_kernel,
+    make_sum_reduction_2d_kernel,
+)
+
+zero_idx = np.int64(0)
+count_one_as_a_long = np.int64(1)
+count_one_as_an_int = np.int32(1)
+
+
+def topk(array_in, k, group_sizes=None):
```
fcharras, line 31: insist on the non-deterministic results in the docstring: the garantee to return valid topks, but not necessarily the same if the threshold value have several occurences

jjerphan (approved):
After an IRL review with @fcharras, this LGTM.

This is a neat algorithm implementation. :100:

Subsequent work includes extensions to support:
 - `argkmin` paving the way to k-NN implementations (does walking bit letters from right to left and flipping the inequalities for values selection suffice?)
 - little endian representations if it makes sense (do we have to extend the mapping of bits for the alphabetic ordering for this case?)

sklearn_numba_dpex/kmeans/tests/test_kmeans.py
fcharras: it's a mistake I think ? i thought it tests the function that has been replaced by our topk :facepalm:

sklearn_numba_dpex/common/topk.py
```diff
@@ -0,0 +1,692 @@
...
+@lru_cache
+def _make_create_radix_histogram_kernel(
+    n_items,
+    work_group_size,
+    sub_group_size,
+    global_mem_cache_size,
+    counts_private_copies_max_cache_occupancy,
+    dtype,
+):
+    check_power_of_2(sub_group_size)
```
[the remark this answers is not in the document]
fcharras: Actually, it seems that bitwise operators abstract endianness and will work in all cases ?

scripts/topk.py
```diff
@@ -0,0 +1,639 @@
...
+
+
+if __name__ == "__main__":
+    dtype = np.float32
+    size = 10
+    k = 3
+    array_in = dpt.asarray(np.random.default_rng().random(size), dtype=np.float32)
+    result = topk(array_in, k)
+    print(np.sort(dpt.asnumpy(result)))
+    print(np.sort(dpt.asnumpy(array_in))[-k:])
```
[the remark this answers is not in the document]
fcharras: does this comment still hold ? I favored hand crafted tests to go over all cases rather than random data. I don't think random data would add something new and it's less nice to debug when it fails :-/ (and for that the seed must be set so it's not so interesting to have random)

scripts/topk.py
```diff
@@ -0,0 +1,639 @@
...
+# code that follows defines bijections to unsigned integers for float32 and float64
+# floats, with the following bitwise transformations:
+#    - reinterpret the float as an unsigned integer (i.e the same buffer used to encode
+#      a float is used to encode an unsigned integer. Said differently, first transform
+#      the float to the integer such as the bits used to encode the integer are the
+#      same than the bits used to encode the float)
+#    - if the float item is positive, set the first bit of the unsigned integer to 1
+#    - if the float item is negative, flip all of the bits of the unsigned integer
+#      (i.e set to 1 if bit is 0, else 0)
+#
```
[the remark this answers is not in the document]
fcharras: Do you suggest we should optimize by skipping the uint mapping when floats are known to all be positive

sklearn_numba_dpex/testing/__init__.py
```diff
@@ -19,3 +21,12 @@ def override_attr_context(obj, **attrs):
...
             setattr(obj, attr_name, attr_value)
+
+
+def _assert_array_equal_any(array, list_of_arrays):
+    try:
+        np.testing.assert_array_equal(array, list_of_arrays[0])
+    except AssertionError:
+        if len(list_of_arrays) == 1:
+            raise
+        _assert_array_equal_any(array, list_of_arrays[1:])
```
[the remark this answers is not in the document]
fcharras: Not so smart I think, I think the version I just pushed is cleaner now

sklearn_numba_dpex/common/tests/test_topk.py
```diff
@@ -0,0 +1,91 @@
...
+        (
+            [0, 8, 8, -3, 4, -5, 6, 7, 8, -9, 8],
+            [8, 8, 8],
+            [[1, 2, 8], [1, 2, 10], [2, 8, 10], [1, 8, 10]],
+        ),
+    ],
+)
+def test_topk(
+    array_in, expected_top_3, list_of_expected_top_3_idx, dtype, work_group_size
+):
```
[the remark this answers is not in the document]
fcharras: Yes ~one of the test case with k=1 gives a segfault. Always the same test so it should be easy to debug ~will debug in this PR.

sklearn_numba_dpex/common/tests/test_topk.py
fcharras: Added comments for clarifying that there is no tie break strategy and that the output is not deterministic.

sklearn_numba_dpex/common/topk.py
fcharras: Moved at the top of the file

fcharras:
TY for the review @jjerphan , I applied your suggestions

Sndianness is not an issue at our level, the low level runtime abstracts the endianness so that the bitwise operations always work the way it's written now.

There's still a failing test (with a segfault, but 100% reproducible) ~ will investigate.

scripts/topk.py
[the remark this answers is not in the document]
jjerphan: Coming back to this comment after our review IRL, I think it might be misleading and I would rather not include it.

fcharras:
Fixed the segfault, path is clear !

jjerphan:
@fcharras @ogrisel: This has been opened for a while, should we redo a review on site together?

fcharras:
Up to this point the top-k is 3x slower than CPU np.partition, on a iGPU.

It looks bad but in fact I think it's decent, I can't compare with `dpnp.partition` yet because it segfaults, but I think the ceiling performance on `iGPU` shouldn't be that high :thinking: 

I think the performance on discrete GPUs will be good, but so far my benchmarks on `flex` are heavily affected by https://github.com/IntelPython/numba-dpex/issues/945 and mostly show overhead...

We do have workarounds for the overhead for `KMeans` but it doesn't apply as good to `top_k`, I'm trying to figure a way out of that but it's not trivial :-(

PR #82, opened 2023-01-11
MAINT update dependencies to 2023 releases + numba_dpex 0.19.0

fcharras:
Last remaining TODO: euclidean distances unit test failing on GPU.

PR #83, opened 2023-01-12
Update dockerfile

fcharras:
Auto-merging this ahead of review to make the CI green again

Review can still go forward in https://github.com/soda-inria/sklearn-numba-dpex/pull/82


fcharras:
Removing the unneeded oneapi binaries divides the size of the image in two (from 10GB to 5GB) !!

PR #84, opened 2023-01-13
Fix a bug that caused some kmeans helper kernels to not be correctly cached

fcharras:
Revert some changes in one of the last cleaning commits in  https://github.com/soda-inria/sklearn-numba-dpex/pull/82 Those changes disrupted the caching of some kernels, causing re-compiling at each iteration in kmeans, with a big performance hit.

Also improves the performance of kmeans++ since there were an even older issue with caching in the random module. 

This issue also teach us that bugs in caching can be disastrous and are totally silent and should always require special care.

### DIFF
--- a/sklearn_numba_dpex/common/random.py
+++ b/sklearn_numba_dpex/common/random.py
@@ -158,39 +158,46 @@ def create_xoroshiro128pp_states(n_states, subsequence_start=0, seed=None, devic
     if hasattr(seed, "randint"):
         seed = uint64(seed.randint(0, np.iinfo(np.int64).max - 1))
 
-    seed = uint64(seed)
+    init_xoroshiro128pp_states_kernel = _make_init_xoroshiro128pp_states_kernel(
+        n_states, subsequence_start
+    )
+
+    # Initialization is purely sequential so it will be faster on CPU, if a cpu device
+    # is available make sure to use it.
+    if device is None:
+        device = dpctl.SyclDevice()
+    from_cpu_to_device = False
+    if not device.has_aspect_cpu:
+        try:
+            cpu_device = dpctl.SyclDevice("cpu")
+            from_cpu_to_device = True
+        except dpctl.SyclDeviceCreationError:
+            warnings.warn(
+                "No CPU found, falling back random initiatlization to default device."
+            )
+
+    states = dpt.empty(
+        sh=(n_states, 2),
+        dtype=np.uint64,
+        device=cpu_device if from_cpu_to_device else device,
+    )
+
+    seed = dpt.asarray(
+        [seed], dtype=np.uint64, device=cpu_device if from_cpu_to_device else device
+    )
+
+    init_xoroshiro128pp_states_kernel(states, seed)
+
+    if from_cpu_to_device:
+        return states.to_device(device)
+
+    else:
+        return states
+
+
+@lru_cache
+def _make_init_xoroshiro128pp_states_kernel(n_states, subsequence_start):
     n_states = int64(n_states)
-
-    init_const_1 = np.uint64(0)
-
-    @dpex.kernel
-    def init_xoroshiro128pp_states(states):
-        """
-        Use SplitMix64 to generate an xoroshiro128p state from a uint64 seed.
-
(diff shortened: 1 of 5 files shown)


sklearn_numba_dpex/common/random.py
```diff
@@ -158,39 +158,46 @@ def create_xoroshiro128pp_states(n_states, subsequence_start=0, seed=None, devic
     if hasattr(seed, "randint"):
```
fcharras, line 158: NB: diff is hardly readable but the changes are simple, it consists in moving some block of codes into a separate `_make_...` function that is decorated with `lru_cache`.

sklearn_numba_dpex/common/random.py
```diff
@@ -158,39 +158,46 @@ def create_xoroshiro128pp_states(n_states, subsequence_start=0, seed=None, devic
...
+        device = dpctl.SyclDevice()
+    from_cpu_to_device = False
+    if not device.has_aspect_cpu:
+        try:
+            cpu_device = dpctl.SyclDevice("cpu")
+            from_cpu_to_device = True
+        except dpctl.SyclDeviceCreationError:
+            warnings.warn(
+                "No CPU found, falling back random initiatlization to default device."
+            )
```
jjerphan, line 177: Side-note: This code is also present in KMeans.

Don't we have an utility function for that somewhere?

sklearn_numba_dpex/common/random.py
```diff
@@ -292,32 +301,36 @@ def _xoroshiro128pp_jump(states, state_idx):
...
-    init_xoroshiro128pp_states[1, 1](states)
+        This ensures that manually set small seeds don't result in a predictable
+        initial sequence from the random number generator.
+        """
+        if n_states < one_idx:
+            return
 
-    if from_cpu_to_device:
-        return states.to_device(device)
+        splitmix64_state = init_const_1 ^ seed[zero_idx]
```
jjerphan, line 315: Note: I find moving blocks of code around hardens understandability of the logic change: it is hard for reviewers to know what has changed in this file with respect to the scope of this PR due to this diff which does not change the logic it seems.

To me this is the only line whose logic changed (accessing `seed[zero_idx]` instead if `seed`) but I am not entirely sure. Is that the case? Is swapping those blocks really needed here?

sklearn_numba_dpex/common/random.py
```diff
@@ -158,39 +158,46 @@ def create_xoroshiro128pp_states(n_states, subsequence_start=0, seed=None, devic
...
 
-        # populate the rest of the array
-        for idx in range(one_idx, n_states):
-            # take state of previous generator
-            states[idx, zero_idx] = states[idx - one_idx, zero_idx]
-            states[idx, one_idx] = states[idx - one_idx, one_idx]
-            # and jump forward 2**64 steps
-            _xoroshiro128pp_jump(states, idx)
+    else:
+        return states
```
jjerphan, line 195: Super small nit.
```suggestion
    return states
```

sklearn_numba_dpex/common/random.py
fcharras: Last time we chose to wait for the third occurrence to chose to factor it. Third occurrence happens in https://github.com/soda-inria/sklearn-numba-dpex/pull/81 + the said refacto.

jjerphan: Ah yes!

sklearn_numba_dpex/common/random.py
fcharras: Unfortunately the fix consists in moving those blocks around to a place where caching can be correctly defined so yes moving the blocks is necessary :-(

Indeed you spotted a difference, seed is now passed as a kernel argument rather than a constant variable to avoid re-triggering the jit each time the seed changes.

jjerphan: Fine for me. :)

jjerphan (approved):
LGTM. I do not want my nitpick to block. :slightly_smiling_face:

fcharras:
Thanks!! :fireworks:

PR #85, opened 2023-01-18
Fix issues seen on devcloud with flex170 gpu

fcharras:
- pin numba_dpex build from dppy/label/dev conda channel
- use a default hardcoded value when device global memory cache size can't be fetched

For the latter fix it will be interesting to test during the next session on the devcloud if clinfo also fails at finding `global_memory_cache_size`, else we could use `pyopencl`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,8 @@ conda create --yes --name $CONDA_DPEX_ENV_NAME \
              --channel dppy/label/dev \
              --channel conda-forge \
              --channel intel \
-             # NB: different versions of `sklearn_numba_dpex` can require to pin
-             # different versions, build or channels here.
+             `# NB: different versions of `sklearn_numba_dpex` can require to pin` \
+             `# different versions, builds or channels here.` \
              numba-dpex=0.19.0=py39hfc4b9b4_5 "intel::dpcpp_linux-64"
 ```
 
--- a/README.md
+++ b/README.md
@@ -119,9 +119,9 @@ conda create --yes --name $CONDA_DPEX_ENV_NAME \
              --channel dppy/label/dev \
              --channel conda-forge \
              --channel intel \
-             `# TODO: remove <2023.0.0 pin after dppy/label/dev is updated` \
-             `# see https://github.com/IntelPython/dpctl/issues/1022` \
-             numba-dpex "intel::dpcpp_linux-64"
+             # NB: different versions of `sklearn_numba_dpex` can require to pin
+             # different versions, build or channels here.
+             numba-dpex=0.19.0=py39hfc4b9b4_5 "intel::dpcpp_linux-64"
 ```
 
 (where you can replace the name of the environment `my-dpex-env` with a name of your
--- a/sklearn_numba_dpex/common/_utils.py
+++ b/sklearn_numba_dpex/common/_utils.py
@@ -1,4 +1,5 @@
 import math
+import warnings
 
 
 def check_power_of_2(e):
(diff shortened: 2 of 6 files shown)


sklearn_numba_dpex/common/_utils.py
```diff
@@ -88,3 +89,25 @@ def _check_max_work_group_size(
         )
     else:
         return work_group_size
+
+
+_GLOBAL_MEM_CACHE_SIZE_DEFAULT = 1048576
```
jjerphan, line 94: Is it possible to document the choice of this value?

Also I think it is worth making the power of 2 explicit
```suggestion
_GLOBAL_MEM_CACHE_SIZE_DEFAULT = 1048576  # 2**20
```

Or even this is clearer to me:

```suggestion
_GLOBAL_MEM_CACHE_SIZE_DEFAULT = 2**20
```

What do you think?

sklearn_numba_dpex/common/_utils.py
```diff
@@ -88,3 +89,25 @@ def _check_max_work_group_size(
...
+    if (global_mem_cache_size := device.global_mem_cache_size) > 0:
+        return global_mem_cache_size
+
+    warnings.warn(
+        "Can't inspect the available global memory cache size for the device "
+        f"{device.name}. Please check that your drivers and runtime libraries are up "
+        "to date, if this warning persists please report it at "
+        "https://github.com/soda-inria/sklearn-numba-dpex/issues . The execution will "
+        "continue with a default value for the cache size set to "
+        f"{_GLOBAL_MEM_CACHE_SIZE_DEFAULT} , which is assumed to be safe but might "
```
jjerphan, line 108:
```suggestion
        "continue with a default value for the cache size set to "
        f"{_GLOBAL_MEM_CACHE_SIZE_DEFAULT} bytes, which is assumed to be safe but might "
```

README.md
```diff
@@ -119,9 +119,9 @@ conda create --yes --name $CONDA_DPEX_ENV_NAME \
              --channel dppy/label/dev \
              --channel conda-forge \
              --channel intel \
-             `# TODO: remove <2023.0.0 pin after dppy/label/dev is updated` \
-             `# see https://github.com/IntelPython/dpctl/issues/1022` \
-             numba-dpex "intel::dpcpp_linux-64"
+             # NB: different versions of `sklearn_numba_dpex` can require to pin
+             # different versions, build or channels here.
```
jjerphan, line 123: Don't we need to escape the comment here like done previously?
```suggestion
             `# NB: different versions of sklearn_numba_dpex can require to pin` \
             `# different channels or versions and builds numbers of dependencies` \
             `# here.` \
```

fcharras:
TY @jjerphan I've applied your suggestions

I also just noticed while updating the TopK code, that I put wrong numbers in the predicted local memory usage of the kmeans kernels (kmeans++, lloyd and compute_labels), which is needed to run on CPU. 

When fixing it I also saw that sometimes SYCL refuses the kernel call if the predicted local memory is too close from the maximum amount of available local memory, so I also ensured to leave a free buffer. (probably that the device itself use it to optimize and load things)

It's a small diff so I took the liberty to hijack it into the last commit.

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -20,17 +20,17 @@ def make_label_assignment_fixed_window_kernel(
     n_samples, n_features, n_clusters, sub_group_size, work_group_size, dtype, device
 ):
     window_n_centroids = sub_group_size
-    centroids_window_width = window_n_centroids + 1
```
fcharras, line 23: Adding +1 here was random try of a trick around bank conflict. But after reading some more on the subject of bank conflicts, I think it makes no sense in this kernel. It doesn't show any performance difference on benchmarks either. Let's just remove it, it also overcomplicated the prediction on local memory usage.

sklearn_numba_dpex/common/_utils.py
```diff
@@ -77,7 +79,11 @@ def _check_max_work_group_size(
         return device.max_work_group_size
     elif work_group_size == "max":
         return math.floor(
-            (device.local_mem_size - required_memory_constant)
+            (
+                device.local_mem_size
```
jjerphan, line 83: Can we be sure that `device.local_mem_size` always stores a valid value?

sklearn_numba_dpex/common/_utils.py
```diff
@@ -88,3 +94,27 @@ def _check_max_work_group_size(
         )
     else:
         return work_group_size
+
+
+# This is the value found for Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics]
+# GPU.
+_GLOBAL_MEM_CACHE_SIZE_DEFAULT = 1048576  # 2**20
```
jjerphan, line 101: Thanks!

jjerphan (approved):

fcharras:
TY for review @jjerphan

PR #86, opened 2023-01-20
FIX make test_error_raised_on_invalid_group_sizes pass on smaller GPUs

ogrisel:
My `max_work_group_size` is 256.

```python
>>> import dpctl
>>> dpctl.SyclDevice()
<dpctl.SyclDevice [backend_type.opencl, device_type.gpu,  Intel(R) Iris(R) Graphics 540 [0x1926]] at 0x7ff4e77fc5f0>
```


jjerphan (approved):

fcharras:
Thanks. If it's still not enough for some devices, we can push it further and write a function that uses device.max_work_group_size to find automatically an appropriate number.

PR #87, opened 2023-01-20
Improve understandability of the sum reduction kernels

ogrisel:
Here is my attempt to make the sum reduction kernel related code more intuitive to a newcomer. It's a mix of change of variable name, order of code blocks and improvements to the docstrings / comments.

I still have follow-up work to do but I am not sure when I will have the time to continue this work, so feel free to edit / improve / merge as you wish without waiting for me.

### DIFF
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -221,7 +221,7 @@ def make_sum_reduction_2d_kernel(
     This implementation is optimized for C-contiguous arrays.
 
     numba_dpex does not provide tools such as `cuda.reduce` so we implement
-    from scratch a reduction strategy. The strategy relies on the associativoty
+    from scratch a reduction strategy. The strategy relies on the associativity
     of the operation used for the reduction, thus allowing to reduce the input
     in any order.
 
@@ -263,8 +263,8 @@ def make_sum_reduction_2d_kernel(
     https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html # noqa
 
     It is expected to take one scalar argument and returning one
-    scalar value. lambda functions are advised against since there compilation
-    cannot be cached.
+    scalar value. lambda functions are advised against since their compilation
+    might not be cached.
 
     `sklearn_numba_dpex.common._utils` exposes some pre-defined functions
     suitable to be passed as `fused_elementwise_func`,
--- a/sklearn_numba_dpex/common/kernels.py
+++ b/sklearn_numba_dpex/common/kernels.py
@@ -216,46 +216,70 @@ def make_sum_reduction_2d_kernel(
     sub_group_size=None,
     fused_elementwise_func=None,
 ):
-    """Implement data_2d.sum(axis=axis) or data_1d.sum().
+    """Compute data_2d.sum(axis=axis) or data_1d.sum().
 
     This implementation is optimized for C-contiguous arrays.
 
-    numba_dpex does not provide tools such as `cuda.reduce` so we implement from scratch
-    a reduction strategy. The strategy relies on the commutativity of the operation used
-    for the reduction, thus allowing to reduce the input in any order.
+    numba_dpex does not provide tools such as `cuda.reduce` so we implement
+    from scratch a reduction strategy. The strategy relies on the associativoty
+    of the operation used for the reduction, thus allowing to reduce the input
+    in any order.
 
-    The strategy consists in performing local reductions in each work group using local
-    memory where each work item combine two values, thus halving the number of values,
-    and the number of active work items. At each iteration the work items are discarded
-    in a bracket manner. The work items with the greatest ids are discarded first, and
-    we rely on the fact that the remaining work items are adjacents to optimize the RW
-    operations.
+    The strategy consists in performing a series of kernel invocations that
+    each perform a partial sum. At each kernel invocation, the input array is
+    tiled with non-overlapping windows and all the values within a given window
+    are collaboratively summed by the work items of a given work group.
 
-    Once the reduction is done in a work group the result is written in global memory,
-    thus creating an intermediary result whose size is divided by
-    `2 * work_group_size` if axis=1, or by `2 * work_group_size // sub_group_size` if
-    axis=0. This is repeated as many times as needed until only one value remains in
-    global memory.
+    Each window has shape:
 
-    If `fused_elementwise_func` is not None, it will be applied element-wise before
-    summing. It must be a function that will be interpreted as a dpex.func and is
-    subject to the same rules ( see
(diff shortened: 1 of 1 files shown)


sklearn_numba_dpex/common/kernels.py
```diff
@@ -330,24 +351,28 @@ def make_sum_reduction_2d_kernel(
     # TODO: at some point, the cost of scheduling the kernel is more than the cost of
     # running the reduction iteration. At this point the loop should stop and then a
     # single work item should iterates one time on the remaining values to finish the
-    # reduction ?
+    # reduction?
     kernel = fused_func_kernel
-    while sum_axis_size > 1:
-        sum_axis_size = math.ceil(sum_axis_size / reduction_block_size)
+    sum_axis_size = size0 if axis == 0 else size1
+    next_input_size = sum_axis_size
```
ogrisel, line 357: Not that introducing a new `next_input_size` to be updated in the loop instead of updating `sum_axis_size` is both helpful to understand the code but also to avoid introducing potential bugs in subsequent code such as the `sum_reductions` local function. At the moment it was only checking for `if sum_axis_size == 0:` which is a special case that was not impacted but if this code evolves it could have introduced weird bugs.

ogrisel:
BTW I checked that the changes in this PR do not impact the performance using the benchmark script.

jjerphan:
Thank you, @ogrisel.

I am currently busy, I let @fcharras review and merge.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -216,46 +216,70 @@ def make_sum_reduction_2d_kernel(
...
+
     https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html # noqa
-    ). It is expected to take one scalar argument and returning one scalar value.
-    lambda functions are advised against since the cache will not work with lambda
-    functions. sklearn_numba_dpex.common._utils expose some pre-defined
-    `fused_elementwise_func`s.
+
+    It is expected to take one scalar argument and returning one
+    scalar value. lambda functions are advised against since there compilation
+    cannot be cached.
```
fcharras, line 267: In fact the cache will work as long as the hash taken by `lru_cache` from the signature is the same. Two different lambda functions, even if they define the same function, will have different hashes (probably hashes with `id(func)` ?).  So the cache will work if a given lambda function is defined once and re-used, and stop working as soon as it's re-defined. :shrug: not sure it's worth mentioning since I'd imagine cache failing in this case can be expected.

So maybe be more vague:

```suggestion
    might not be cached.
```

fcharras (changes requested):
TY for contributing to the effort at clarity in this repo @ogrisel . I think `associativity` is not enough (see comments below). Left two other comments not relevant for approval.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -216,46 +216,70 @@ def make_sum_reduction_2d_kernel(
...
-    thus creating an intermediary result whose size is divided by
-    `2 * work_group_size` if axis=1, or by `2 * work_group_size // sub_group_size` if
-    axis=0. This is repeated as many times as needed until only one value remains in
-    global memory.
-
-    If `fused_elementwise_func` is not None, it will be applied element-wise before
-    summing. It must be a function that will be interpreted as a dpex.func and is
-    subject to the same rules ( see
+    numba_dpex does not provide tools such as `cuda.reduce` so we implement
+    from scratch a reduction strategy. The strategy relies on the associativoty
```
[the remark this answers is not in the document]
ogrisel: Fair enough.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -216,46 +216,70 @@ def make_sum_reduction_2d_kernel(
...
-    Once the reduction is done in a work group the result is written in global memory,
-    thus creating an intermediary result whose size is divided by
-    `2 * work_group_size` if axis=1, or by `2 * work_group_size // sub_group_size` if
-    axis=0. This is repeated as many times as needed until only one value remains in
-    global memory.
-
-    If `fused_elementwise_func` is not None, it will be applied element-wise before
-    summing. It must be a function that will be interpreted as a dpex.func and is
-    subject to the same rules ( see
+    numba_dpex does not provide tools such as `cuda.reduce` so we implement
```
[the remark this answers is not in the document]
ogrisel: In scikit-learn learn we still use 79 char limits for the docstrings (while we use 88 char limit for the code itself).

I did not reconfigure my editor when making those edit so I used the scikit-learn convention here.

Note that strictly following PEP8 would require 72 chars for the docstrings but I think nowadays few projects follow this.

But since this is not scikit-learn, I don't have a strong opinion, feel free to rewrap to 88.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -216,46 +216,70 @@ def make_sum_reduction_2d_kernel(
...
+    limitations as explained in:
+
     https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html # noqa
-    ). It is expected to take one scalar argument and returning one scalar value.
-    lambda functions are advised against since the cache will not work with lambda
-    functions. sklearn_numba_dpex.common._utils expose some pre-defined
-    `fused_elementwise_func`s.
+
+    It is expected to take one scalar argument and returning one
+    scalar value. lambda functions are advised against since there compilation
```
ogrisel, line 266:
```suggestion
    scalar value. lambda functions are advised against since their compilation
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -216,46 +216,70 @@ def make_sum_reduction_2d_kernel(
...
-    axis=0. This is repeated as many times as needed until only one value remains in
-    global memory.
-
-    If `fused_elementwise_func` is not None, it will be applied element-wise before
-    summing. It must be a function that will be interpreted as a dpex.func and is
-    subject to the same rules ( see
+    numba_dpex does not provide tools such as `cuda.reduce` so we implement
+    from scratch a reduction strategy. The strategy relies on the associativity
+    of the operation used for the reduction, thus allowing to reduce the input
+    in any order.
```
ogrisel, line 226:
```suggestion
    from scratch a reduction strategy. The strategy relies on the associativity
    and commutativity of the operation used for the reduction, thus allowing to
    reduce the input in any order.
```

sklearn_numba_dpex/common/kernels.py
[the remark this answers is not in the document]
fcharras: It's more that I like to have automated tools that enforce things to avoid future discussions in PRs. Is it something that is enforced by the CI in scikit-learn ?

fcharras:
TY @ogrisel 

Adding the question of automated formatting of docstring (especially width) to a future TODO list.

PR #88, opened 2023-02-06
ENH small cleaning and optimizations accross the repo 

fcharras:
- [x] Replace signature exposing sizes with a signature exposing one argument `shape`
- [x] Where applicable, use 2D grid of work items rather than 1D + divisions 
    + note: found this piece of information: https://stackoverflow.com/a/15044884 suggesting that:
        + 2D grid really are better for performance because doing // and % ops in the kernel is expensive
        + it indexes work items in a "row major order" meaning that items in the same row belong to the same sub group (/warp)
    + even if the previous information applies to cuda. I assume it's also true for numba-dpex/ sycl
- [x] some kernels are now written only for a given shape and reused for other dimension with reshaping tricks.
- [x] also use 2D grids for kmeans kernels
- [x] document better the use of `centroids_private_copies_max_cache_occupancy` and improve the heuristic using `device.max_compute_units`
- [x] a couple more nitpicks

~~It's WIP because using 2D groups for the sum over axis 0 kernel seems to trigger weirg bugs like in https://github.com/IntelPython/numba-dpex/issues/892 in some of the tests.~~

Out of WIP, remaining issues were unrelated, PR is green.

edit: confirmed affected by https://github.com/IntelPython/numba-dpex/issues/906


fcharras:
Out of WIP. See the top level post for the list of changes.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -463,25 +461,26 @@ def prepare_data_for_lloyd(X_t, init, tol, sample_weight, copy_x):
...
         work_group_size="max",
         device=device,
         dtype=compute_dtype,
     )
 
-    elementwise_binary_divide_kernel = make_elementwise_binary_op_1d_kernel(
-        n_features, _divide, max_work_group_size, compute_dtype
+    divide_by_n_samples_kernel = make_apply_elementwise_func(
+        (n_features,),
+        _divide_by(compute_dtype(n_samples)),
```
jjerphan, line 473: I do not find this chaining super intuitive. Can you extract it above just above the call to this function and give it a name? :slightly_smiling_face:

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
```diff
@@ -230,5 +234,9 @@ def kmeansplusplus_single_step(
 
             dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
 
-    global_size = (math.ceil(n_samples / work_group_size)) * (work_group_size)
-    return kmeansplusplus_single_step[global_size, work_group_size]
+    global_size = (
+        math.ceil(math.ceil(n_samples / window_n_candidates) / candidates_window_height)
+        * candidates_window_height,
```
jjerphan, line 239: Can you give this a name?

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -106,22 +106,46 @@ def make_lloyd_single_step_fixed_window_kernel(
     last_centroid_window_idx = n_windows_for_centroids - 1
     last_feature_window_idx = n_windows_for_features - 1
 
-    centroids_window_shape = (centroids_window_height, centroids_window_width)
-
-    global_size = (math.ceil(n_samples / work_group_size)) * (work_group_size)
-    n_subgroups = global_size // sub_group_size
+    n_subgroups = math.ceil(n_samples / window_n_centroids)
+    global_size = (
+        math.ceil(n_subgroups / centroids_window_height) * centroids_window_height,
```
jjerphan, line 111: Identically, is it possible to give this a name?

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -161,5 +170,9 @@ def assignment(
 
         assignments_idx[sample_idx] = min_idx
 
-    global_size = (math.ceil(n_samples / work_group_size)) * (work_group_size)
-    return assignment[global_size, work_group_size]
+    global_size = (
+        math.ceil(math.ceil(n_samples / window_n_centroids) / centroids_window_height)
+        * centroids_window_height,
```
jjerphan, line 175: Identically, is it possible to give this a name?

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -334,9 +361,7 @@ def fused_lloyd_single_step(
         # counts.
 
         # each work item is assigned an array of centroids in a round robin manner
-        privatization_idx = (
-            sample_idx // sub_group_size
-        ) % n_centroids_private_copies
+        privatization_idx = sub_group_idx % n_centroids_private_copies
```
jjerphan, line 364: Is it possible to avoid using a modulo, here?

jjerphan (commented):
Thank you for this PR of general improvements, @fcharras. :slightly_smiling_face: 

Here is a first pass.

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
fcharras: I don't think so :-(

sklearn_numba_dpex/kmeans/kernels/kmeans_plusplus.py
fcharras: I agree but I don't really now how to it clear, it's twisted that the shape of the grid is adapted to the sliding window on centroids, but each work item apply to one single input sample.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -760,13 +738,19 @@ def partial_sum_reduction(
...
             work_item_row_idx = first_row_idx + local_row_idx + n_active_sub_groups
             if (
                 (local_row_idx < n_active_sub_groups) and
-                (col_idx < n_cols) and
+                # HACK: this condition save some memory access and was added thinking
+                # it might be good for performance. However on CPU it causes wrong
+                # results in some tests.
+                # TODO: create a minimal reproducer and report to `numba_dpex`. Remove
+                # the hack once it is fixed.
+                (_is_cpu or (col_idx < n_cols)) and
```
fcharras, line 753: There's probably some bug in numba_dpex or with the cpu backend here that causes branching to not be correctly executed in very specific cases, for this very particular condition, on cpu. I had no luck when trying to have a minimal reproducer and I can't see failures on gpu. So I'd be in favor of letting this pass through in the meantime. Note that this also fixes the case with `work_group_size=1` that I added back to the unit tests.

fcharras: Well it's still wrong on the CI. :shrug:

fcharras:
Something is wrong with `sum(axis=1)` on CPU when `work_group_size` is `>=8`. Maybe another bug in the JIT or something we don't understand about group sizes on CPU (or a bit of both). In both cases the investigation looks complicated and CPU is not the priority target so the latest commit propose forcing `work_group_size == 8` with `work_group_size` was set to `max`.

I'll also run this branch on the dev cloud with the flex170 to see if it shows any performance improvement, and play with the group size to see if it affects performance more than it seems to do on local iGPU.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -610,39 +579,58 @@ def fused_elementwise_func_(x):
...
     else:
         # Round work_group_size to the maximum smaller power-of-two multiple of
         # `sub_group_size`
         n_sub_groups_per_work_group = get_maximum_power_of_2_smaller_than(
             work_group_size / sub_group_size
         )
+        n_sub_groups_per_work_group = n_sub_groups_per_work_group
         work_group_size = n_sub_groups_per_work_group * sub_group_size
 
+    _is_cpu = device.has_aspect_cpu
```
ogrisel, line 594: Nit: drop the leading `_` in the variable name if you don't expect a name conflict.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -610,39 +579,58 @@ def fused_elementwise_func_(x):
...
+            reduction_block_size,
+        ),
+        (get_result_shape, get_global_size),
+    )  # shape_update_fn
+
 
-@lru_cache
 def _make_partial_sum_reduction_2d_axis0_kernel(
-    n_cols, work_group_size, sub_group_size, fused_elementwise_func, dtype
+    n_cols, work_group_size, sub_group_size, fused_elementwise_func, dtype, _is_cpu
```
ogrisel, line 633: Same comment here.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -728,13 +707,13 @@ def partial_sum_reduction(
...
-        local_values = dpex.local.array(local_values_size, dtype=dtype)
+        local_values = dpex.local.array(work_group_shape, dtype=dtype)
 
         # The current work item use the following second coordinate (given by the
         # position of the window in the grid of windows, and by the local position of
         # the work item in the 2D index):
         col_idx = (
-            (group_id // n_blocks_per_col) * sub_group_size + local_col_idx
+            (dpex.get_group_id(one_idx) * sub_group_size) + local_col_idx
             )
```
ogrisel, line 717: black:

```suggestion
        )
```

which makes me think that I wanted to open a PR to fix the black problem but completely forgot about it. I will wait for this PR to me finalized first though.

sklearn_numba_dpex/kmeans/kernels/_base_kmeans_kernel_funcs.py
```diff
@@ -229,10 +235,9 @@ def _initialize_window_of_centroids(
             # The first `window_n_centroids` work items cooperate on loading the
             # values of centroids_half_l2_norm relevant to current window. Each work
             # item loads one single value.
-            if local_work_id < window_n_centroids:
-                half_l2_norm_loading_idx = first_centroid_idx + local_work_id
-                window_of_centroids_half_l2_norms[local_work_id] = (
-                    centroids_half_l2_norm[half_l2_norm_loading_idx]
+            if local_row_idx == zero_as_a_long:
```
ogrisel, line 238: Are you sure this change is correct? Don't we need a bound check on the value of `local_col_idx`? I would have maybe expected:

```suggestion
            if local_row_idx == zero_as_a_long and local_col_idx < window_n_centroids:
```

assuming that `window_of_centroids_half_l2_norms` is raveled window of logical shape `(1, window_n_centroids)`, but the comment should probably be updated to better explain the interaction between `local_row_idx`, `local_col_idx` and the shape of the window array.

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -80,9 +85,12 @@ def assignment(
...
         # fmt: on
-
-        sample_idx = dpex.get_global_id(zero_idx)
-        local_work_id = dpex.get_local_id(zero_idx)
+        local_row_idx = dpex.get_local_id(zero_idx)
+        local_col_idx = dpex.get_local_id(one_idx)
+        sample_idx = (
+            (dpex.get_global_id(zero_idx) * sub_group_size)
+            + local_col_idx
+        )
```
ogrisel, line 93: I admit I am lost here. Could you add some comments to explain how `sub_group_size` and the 2d local indexing logic work?

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -36,12 +36,16 @@ def make_label_assignment_fixed_window_kernel(
...
-    elif centroids_window_height * sub_group_size != work_group_size:
+    if (work_group_size == input_work_group_size) and (
+        (centroids_window_height * sub_group_size) != work_group_size
+    ):
         raise ValueError(
             "Expected work_group_size to be a multiple of sub_group_size but got "
             f"sub_group_size={sub_group_size} and work_group_size={work_group_size}"
         )
 
+    work_group_shape = (centroids_window_height, window_n_centroids)
```
ogrisel, line 47:
```suggestion
    work_group_shape = (centroids_window_height, centroids_window_width)
```

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
```diff
@@ -27,7 +27,7 @@ def make_label_assignment_fixed_window_kernel(
         work_group_size,
         device,
         required_local_memory_per_item=dtype_itemsize,
-        required_memory_constant=sub_group_size * dtype_itemsize,
+        required_memory_constant=window_n_centroids * dtype_itemsize,
     )
 
     centroids_window_width = window_n_centroids
```
ogrisel, line 33: Maybe we could do:

```suggestion
    centroids_window_width = sub_group_size
```

and completely drop the `window_n_centroids` variable to remove on indirection?

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -106,22 +106,46 @@ def make_lloyd_single_step_fixed_window_kernel(
...
 
     n_cluster_items = n_clusters * (n_features + 1)
     n_cluster_bytes = np.dtype(dtype).itemsize * n_cluster_items
     # TODO: control that this value is not higher than the number of sub-groups of size
     # sub_group_size that can effectively run concurrently. We should fetch this
     # information and apply it here.
+
+    # NB: for more details about the privatization strategy the following variables
+    # refer too, please read the inline commenting that address it with more details in
+    # the kernel definition.
```
ogrisel, line 123:
```suggestion
    # refer too, please read the inline comments in the kernel definition.
```

to avoid duplicating "more details".

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -106,22 +106,46 @@ def make_lloyd_single_step_fixed_window_kernel(
...
+    # Ensure that the memory allocated for privatization will not saturate the global
+    # memory cache size. For this purpose we limit the number of private copies to
+    # a fraction of the available global memory cache size.
     n_centroids_private_copies = (
         global_mem_cache_size * centroids_private_copies_max_cache_occupancy
     ) // n_cluster_bytes
-    n_centroids_private_copies = int(min(n_subgroups, n_centroids_private_copies))
+
+    # Each set of `sub_group_size` consecutive work items is assigned one private
+    # copy, and several such set can be assigned to the same private copy. Thus, at
```
ogrisel, line 133:
```suggestion
    # copy, and several such sets can be assigned to the same private copy. Thus, at
```

sklearn_numba_dpex/kmeans/kernels/_base_kmeans_kernel_funcs.py
```diff
@@ -217,7 +222,8 @@ def _initialize_results(results):
         @dpex.func
         # fmt: off
         def _initialize_window_of_centroids(
-            local_work_id,                      # PARAM
+            local_row_idx,                      # PARAM
+            local_col_idx,                      # PARAM
             first_centroid_idx,                 # PARAM
             centroids_half_l2_norm,             # IN
             window_of_centroids_half_l2_norms,  # OUT
```
ogrisel, line 229: Making the shapes explicit after IN / OUT would probably help.

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -199,45 +199,40 @@ def make_reduce_centroid_data_kernel(
...
-        centroids_t[feature_idx, cluster_idx] = sum_
 
-        # reduce the cluster sizes
-        if feature_idx == zero_idx:
-            sum_ = zero
+        # reduce the centroid values
+        if item_idx < n_centroid_items:
+            for copy_idx in range(n_centroids_private_copies):
+                sum_ += centroids_t_private_copies[copy_idx, item_idx]
+                centroids_t[item_idx] = sum_
```
ogrisel, line 232: Isn't there an indentation problem here?

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -199,45 +199,40 @@ def make_reduce_centroid_data_kernel(
...
+
+        elif item_idx < n_sums:
```
ogrisel, line 234: This seems redundant with the `if item_idx >= n_sums:` early return at the beginning of the kernel.

sklearn_numba_dpex/kmeans/kernels/utils.py
```diff
@@ -199,45 +199,40 @@ def make_reduce_centroid_data_kernel(
...
     @dpex.kernel
     # fmt: off
-    def reduce_centroid_data(
+    def _reduce_centroid_data_kernel(
         cluster_sizes_private_copies,  # IN      (n_copies, n_clusters)
-        centroids_t_private_copies,    # IN      (n_copies, n_features, n_clusters)
+        centroids_t_private_copies,    # IN      (n_copies, n_features * n_clusters) (reshaped)  # noqa
         cluster_sizes,                 # OUT     (n_clusters,)
-        centroids_t,                   # OUT     (n_features, n_clusters)
+        centroids_t,                   # OUT     (n_features * n_clusters,) (flattened)
```
ogrisel, line 217: Maybe rename to `centroids_t_flattened`

ogrisel (commented):
Overall, the changes improve the readability a lot (I think). Looking forward to the performance impacts (once the cause of the bug has been found).

I have seen a bunch of fishing things (see some of the comments below). Maybe this can explain the broken tests?

fcharras:
I've reduced the last failure on the pipeline to what appear to be a JIT issue. The merge will be blocked until it's fixed. It's likely the same JIT issue we've seen before and it gave a much nicer reproducer. See https://github.com/IntelPython/numba-dpex/issues/906 .

sklearn_numba_dpex/common/kernels.py
```diff
@@ -300,46 +270,40 @@ def make_sum_reduction_2d_kernel(
...
-        get_global_size = (
-            lambda result_sum_axis_size: result_sum_axis_size * work_group_size * size0
+        work_group_shape, kernels, shape_update_fn = _prepare_sum_reduction_2d_axis1(
+            shape0, work_group_size, fused_elementwise_func, dtype, device
         )
 
     # XXX: The kernels seem to work fine with work_group_size==1 on GPU but fail on CPU.
-    if work_group_size == 1:
+    if math.prod(work_group_shape) == 1:
         raise NotImplementedError("work_group_size==1 is not supported.")
```
jjerphan, line 296: Should the message be changed here?

sklearn_numba_dpex/common/kernels.py
```diff
@@ -429,17 +395,26 @@ def fused_elementwise_func_(x):
...
     )
 
+    return (
+        work_group_shape,
+        (
+            (partial_sum_reduction, partial_sum_reduction_nofunc),  # kernels
+            reduction_block_size,
+        ),
+        (get_result_shape, get_global_size),
+    )  # shape_update_fn
```
jjerphan, line 415: The last comment likely is off due to black formatting.

What do you think of this suggestion?
```suggestion
    kernels = (partial_sum_reduction, partial_sum_reduction_nofunc)
    shape_update_fn = (get_result_shape, get_global_size)
    return (
        work_group_shape,
        (
            kernels,
            reduction_block_size,
        ),
        shape_update_fn,
    )
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -610,39 +579,58 @@ def fused_elementwise_func_(x):
...
```
jjerphan, line 629: The previous suggestion also applies here.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -760,14 +739,15 @@ def partial_sum_reduction(
...
             # are discarded.
             n_active_sub_groups = n_active_sub_groups // two_as_a_long
             work_item_row_idx = first_row_idx + local_row_idx + n_active_sub_groups
             if (
                 (local_row_idx < n_active_sub_groups) and
-                (col_idx < n_cols) and
-                (work_item_row_idx < sum_axis_size)
+                (_is_cpu or ((col_idx < n_cols) and
+                             (work_item_row_idx < sum_axis_size)))
             ):
```
jjerphan, line 751: Can you add inline comment to give some precisions regarding this branch?

jjerphan (approved):
LGTM once @ogrisel's comment have been addressed.

sklearn_numba_dpex/common/kernels.py
```diff
@@ -47,80 +38,59 @@
...
 
-    return initialize_to_zeros[global_size, work_group_size]
+    return initialize_to_zeros
 
 
 @lru_cache
-def make_broadcast_division_1d_2d_axis0_kernel(size0, size1, work_group_size):
-    global_size = math.ceil(size1 / work_group_size) * work_group_size
+def make_broadcast_division_1d_2d_axis0_kernel(shape, work_group_size):
+    n_rows, n_cols = shape
```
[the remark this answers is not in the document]
fcharras: The intent here was to mean that the broadcasted data is `1d`. Would `1d_to_2d` make it clearer ?

sklearn_numba_dpex/common/kernels.py
fcharras: I'll revert this `_is_cpu` change, in fact it does not fix the JIT issue. We must wait for the root cause to be addressed in `numba_dpex`

sklearn_numba_dpex/kmeans/kernels/_base_kmeans_kernel_funcs.py
fcharras: Well spotted, I think your suggestion is right. There is a risk to read `centroids_half_l2_norm` array in global memory outside of its boundaries. Not sure why execution on cpu doesn't segfaults :thinking: 

My guess is that the compiler detected that the corresponding write in local memory is not used later on (which is True) and optimized in a way that skipped the risk of segfault.

fcharras: Will update comment accordingly

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
fcharras: Will add a dosctring in the `lloyd` kernel (where most of the explanations for similar kernels are located)

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
fcharras: or let's just drop `centroids_window_width` variable ? (that was my choice for other files, I probably forgot this one)

sklearn_numba_dpex/kmeans/kernels/compute_labels.py
fcharras: I propose removing `centroids_window_width` completely here. (see previous comment)

Now, regarding using `sub_group_size` or `window_n_centroids` I don't know what is better. In the current iteration I've favored using `window_n_centroids`.

sklearn_numba_dpex/kmeans/kernels/utils.py
fcharras: good catch :relieved: TY

sklearn_numba_dpex/common/kernels.py
fcharras: In my mind `work_group_size=math.prod(work_group_shape)` so I think it's good.

There is this nuance because I have introduced the syntax `work_group_size=max` where we try to automatically use the maximum possible value for `work_group_size`. This change enable checking that `max` does not result in `work_group_size=1`.

sklearn_numba_dpex/common/kernels.py
fcharras: I'll revert this `_is_cpu` change, in fact it does not fix the JIT issue. We must wait for the root cause to be addressed in `numba_dpex`

fcharras:
TY for the careful review @ogrisel @jjerphan the last commit should address your suggestions and I've answered some of your comments.

jjerphan:
Perfect. :+1: 

I leave @ogrisel merge.

ogrisel:
The tests are still red. Ok for merge once fixed.

fcharras:
The tests that fail because of https://github.com/IntelPython/numba-dpex/issues/906 we can merge as soon as it's fixed and we can bump

If there's a conflict with other branches before that we can merge except the diff on the 2d sum kernel.

fcharras:
Closed via https://github.com/soda-inria/sklearn-numba-dpex/pull/98 and https://github.com/soda-inria/sklearn-numba-dpex/pull/96

PR #89, opened 2023-02-08
FEAT: implement convert to sklearn types

fcharras:
To review after

- [x] https://github.com/soda-inria/sklearn-numba-dpex/pull/74 is merged


jjerphan (approved):

fcharras:
I reverted the second commit, it makes the pipeline fail and I think the case is more subtle than what we've discussed. I'll re-open it in a separate PR. I propose merging only the first commit.  (note that the two commits were independant anyway so that @jjerphan approval should still allow going forward)

fcharras:
Merging with follow-up on issues regarding fallback in scikit learn unit tests here: https://github.com/soda-inria/sklearn-numba-dpex/pull/91

PR #90, opened 2023-02-13
ENH Add CLI to the benchmark script.

fcharras:


jjerphan:
Hi @fcharras, is this PR reviewable?

fcharras:
Yes @jjerphan

benchmark/kmeans.py
```diff
@@ -169,35 +259,163 @@ def _check_same_fit(self, estimator, name, max_iter, assert_allclose):
...
+            "Only run the selected engine providers. If not specified, include all "
+            "providers in the benchmark."
+        ),
+    )
+
+    argparser.add_argument(
+        "--include-slow",
+        action="store_true",
+        help="Include in the benchmark the engine providers that are known to be slow.",
+    )
```
jjerphan, line 353: Why do not we use `skip_slow`?

jjerphan (approved):
LGTM. Here are just nitpicks.

benchmark/kmeans.py
fcharras: Because we want `skip_slow` to be true by default, and `--include-slow` rings nicer than `--skip-slow False` or `--not-skip-slow`

PR #93, opened 2023-02-21
MNT bump dependencies and update, including removing hacks required by 0.19.0

fcharras:
Tests also seems affected by https://github.com/IntelPython/numba-dpex/issues/906

Will remain WIP until then


fcharras:
The pipeline can only pass after the merge on main triggers refresh of the docker image with the bumps. Merging now, will re-trigger the pipelines once the docker image is pushed.

PR #96, opened 2023-02-27
Fix work around #906

fcharras:
We were having more and more occurences of https://github.com/IntelPython/numba-dpex/issues/906 to the point that it felt counter-productive to continue developing  with `numba_dpex` until it's fixed.

After torturing the reproducers it seems I've found a way to consistently work around the bug, without having to change the logic of the kernels.

All kernels that contain barriers seems to be affected. The workaround basically consists in moving instructions that contain `array.__setitem__` calls (or atomic functions) in those kernels to `dpex.func` device functions.

The added unit test in this PR demonstrate how it works on a small example. I will also expand on it in https://github.com/IntelPython/numba-dpex/issues/906 .

The minimal reproducer show that not _all_ such instructions seems to require to be moved to device functions, only replacing a select few ones seems to be enough to make the bugs disappear, but since the trigger condition remain unclear for this bug I went the safe way and replaced them all.

The downside is of course readability, but I don't think we should try to improve it now (having the constraint of incorporating those unnatural `dpex.func` factorizations make it very challenging anyway) let's rather move on and hope the bug is fixed upstream in the coming month so we can revert all the hacks.

Three more unrelated fixes have infiltrated the PR:
- [x] A dtype mismatch that fails some sklearn tests when the gpu does not support float64 compute
- [x] I noticed that our tolerance threshold was severely off for large datasets (we forgot to divide it by `n_samples`...), not sure how we let this slip through
- [x] On my machine, the `scikit-learn` test `test_kmeans_verbose[0-lloyd]` suddenly decided that from now on it should fail half of the time. Its reason is a mechanism I'm skeptical about in `scikit-learn` that is called "strict convergence" that is met by comparing labels that have been computed by two successive iterations of lloyd. I had to implement the exact same behavior in our `lloyd` to consistently pass the test, but I think it should only be done when `tol == 0` (which is enough to pass the tests), see arguments in an inline comment in the PR.


sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -313,14 +318,57 @@ def lloyd(
...
+        # this check be in general greater than what the benefits ?
+
+        # Following this reasoning, unlike scikit-learn, we choose to enforce this
+        # behavior if and only if `tol == 0`, because in this case, it is easy to see
+        # that lloyd can indeed fail to stop at the right time due to numerical errors.
+        # Moreover, this is enough to pass scikit-learn unit tests. When `tol > 0`, we
+        # rely on the user setting an appropriate tolerance threshold.
+
+        # TODO: open an issue at `scikit-learn` and propose to adopt this behavior
+        # instead ?
```
ogrisel, line 339:
```suggestion
        # See: https://github.com/scikit-learn/scikit-learn/issues/25716
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -313,14 +318,57 @@ def lloyd(
...
+        # check for strict convergence at each iteration, but that sounds expensive
+        # since it requires an additional pass on each sample label every time.
+        # Providing the user chooses a sensible value for `tol`, wouldn't the cost of
+        # this check be in general greater than what the benefits ?
+
+        # Following this reasoning, unlike scikit-learn, we choose to enforce this
+        # behavior if and only if `tol == 0`, because in this case, it is easy to see
+        # that lloyd can indeed fail to stop at the right time due to numerical errors.
+        # Moreover, this is enough to pass scikit-learn unit tests. When `tol > 0`, we
+        # rely on the user setting an appropriate tolerance threshold.
```
ogrisel, line 336: I would rather track the exact behavior of scikit-learn w.r.t. convergence checks.

sklearn_numba_dpex/patches/tests/test_patches.py
```diff
@@ -86,3 +87,82 @@ def test_spirv_fix():
                 kmeans.fit(X_array)
         finally:
             _load_numba_dpex_with_patches()
+
+
+def test_hack_906():
```
ogrisel, line 92: Let's use an explicit function name:

```suggestion
def test_need_to_workaround_numba_dpex_906():
```

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
```diff
@@ -118,18 +118,39 @@ def compute_distances(
...
             dpex.barrier(dpex.CLK_LOCAL_MEM_FENCE)
 
+    # HACK 906: see sklearn_numba_dpex.patches.tests.test_patches.test_hack_906
+    @dpex.func
+    # fmt: off
+    def _save_distance(
+        sample_idx,                 # PARAM
+        first_centroid_idx,         # PARAM
+        euclidean_distances_t,      # IN
+        sq_distances                # OUT
```
ogrisel, line 140: Please swap the last two args as `sq_distances` is the input and `euclidean_distances_t` is the output.

ogrisel (approved):
LGTM. Just some nits:

sklearn_numba_dpex/kmeans/kernels/compute_euclidean_distances.py
fcharras: good catch, TY

sklearn_numba_dpex/common/kernels.py
```diff
@@ -633,6 +700,9 @@ def _make_partial_sum_reduction_2d_axis0_kernel(
     reduction_block_size = 2 * n_sub_groups_per_work_group
     work_group_shape = (n_sub_groups_per_work_group, sub_group_size)
 
+    local_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
+    global_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
```
jjerphan, line 704: Why do we need two of those and not just one? Is it for this reason?

```suggestion
    # HACK: must define twice to work around the bug highlighted in
    # test_regression_fix
    local_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
    global_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -434,6 +464,9 @@ def _make_partial_sum_reduction_2d_axis1_kernel(
     reduction_block_size = 2 * work_group_size
     work_group_shape = (1, work_group_size)
 
+    local_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
+    global_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
```
jjerphan, line 468: Why do we need two of those and not just one? Is it for this reason?

```suggestion
    # HACK: must define twice to work around the bug highlighted in
    # test_regression_fix
    local_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
    global_sum_and_set_items_if = _make_sum_and_set_items_if_kernel_func()
```

sklearn_numba_dpex/patches/tests/test_patches.py
```diff
@@ -86,3 +87,82 @@ def test_spirv_fix():
...
+    _local_setitem_if = make_setitem_if_kernel_func()
+    _global_setitem_if = make_setitem_if_kernel_func()
+
+    result = dpt.zeros(sh=(1), dtype=dtype, device=dpctl.SyclDevice("cpu"))
+    kernel[32, 32](result)
+
+    assert dpt.asnumpy(result)[0] == 10
+
+
+# HACK 906
```
jjerphan, line 160:
```suggestion
# HACK 906: see sklearn_numba_dpex.patches.tests.test_patches.test_need_to_workaround_numba_dpex_906 # noqa
```

sklearn_numba_dpex/common/kernels.py
```diff
@@ -805,55 +931,127 @@ def partial_argmin_reduction(
...
+
+        local_argmin[local_work_id] = y_idx
+        local_values[local_work_id] = y
 
-        if first_work_id:
-            if local_values[zero_idx] <= local_values[one_idx]:
-                argmin_indices[group_id] = local_argmin[zero_idx]
-            else:
-                argmin_indices[group_id] = local_argmin[one_idx]
+    # HACK 906
```
jjerphan, line 1012:
```suggestion
    # HACK 906: see sklearn_numba_dpex.patches.tests.test_patches.test_need_to_workaround_numba_dpex_906 # noqa
```

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -313,14 +315,55 @@ def lloyd(
         # refers. For this reason, this strategy is not compatible with sklearn
         # unit tests, that consider new_centroids_t (the array after the update)
         # to be the best centroids at each iteration.
-
         centroids_t, new_centroids_t = (new_centroids_t, centroids_t)
 
+        # ???: if two successive assignations have been computed equal, it's called
+        # "strict convergence" and means that the algorithm has converged and can't get
```
jjerphan, line 321: What is the meaning of the `???` here? Is it a question?

jjerphan (approved):
LGTM.

Thank you, @fcharras.

sklearn_numba_dpex/kmeans/drivers.py
```diff
@@ -539,7 +582,7 @@ def prepare_data_for_lloyd(X_t, init, tol, sample_weight, copy_x):
 
     variance = variance_kernel(dpt.reshape(X_t, -1))
     # Use numpy type to work around https://github.com/IntelPython/dpnp/issues/1238
-    tol = (dpt.asnumpy(variance)[0] / n_features) * tol
+    tol = (dpt.asnumpy(variance)[0] / (n_features * n_samples)) * tol
```
jjerphan, line 585: Side-note: Is this change related to the scope of this PR?

I am fine integrating it as part of this PR.

sklearn_numba_dpex/common/kernels.py
fcharras: No need to worry too much about this, the hack will be removed everywhere after bump to `0.20.0dev3`, the PR has been opened https://github.com/soda-inria/sklearn-numba-dpex/pull/93, and hope can be merged soon after this one

sklearn_numba_dpex/kmeans/drivers.py
fcharras: it opens the discsussion regarding diverging from sklearn behavior or considering changing it upstream (https://github.com/scikit-learn/scikit-learn/issues/25716)

sklearn_numba_dpex/kmeans/drivers.py
fcharras: Not related (but a one line fix)

fcharras:
TY for the reviews @jjerphan @ogrisel I've applied your suggestions and answered the question.

I've added an additional commit that fuse strict convergence checking with the main lloyd kernel.

I'll go ahead and merge if it's green after launch.

fcharras:
So I noticed that using 2D work group size harms performance (kernels 30% slower). The last commit revert the 2D work group size diff from https://github.com/soda-inria/sklearn-numba-dpex/pull/88 . I will open afterward a revert PR for this revert commit, maybe if things are fixed in `numba_dpex` regarding 2D group sizes we can merge again later. (theoritically, it should be better)

PR #101, opened 2023-03-01
Move reduction kernels to reduction module

fcharras:
I swear on the Zen of Python that I have only copy/pasted large blocks of code and not maliciously dissimulated opinionated and controversial changes in this PR.


fcharras:
Will merge when test can pass (must wait for https://github.com/soda-inria/sklearn-numba-dpex/actions/runs/4303504742/jobs/7503319872 to be completed first for new docker image to be available)

PR #102, opened 2023-03-01
FEA Implement an optimized sliding window matmul kernel

fcharras:
Necessary step before going forward with the KNN. This kernel will enable computing pairwise distances.

### DIFF
--- a/sklearn_numba_dpex/common/matmul.py
+++ b/sklearn_numba_dpex/common/matmul.py
@@ -1,0 +1,286 @@
+import math
+from functools import lru_cache
+
+import numba_dpex as dpex
+import numpy as np
+
+from sklearn_numba_dpex.common._utils import _check_max_work_group_size
+
+zero_idx = np.int64(0)
+
+
+@lru_cache
+def make_matmul_2d_kernel(
+    X_n_rows,
+    Y_t_n_rows,
+    n_cols,
+    work_group_size,
+    sub_group_size,
+    dtype,
+    device,
+    multiply_fn=None,
+    out_fused_elementwise_fn=None,
+):
+    """Returns a matmul kernel.
+
+    This kernel is about doing a matrix product between `X` and `Y`, but `Y` will be
+    repesented with its transposed `Y_t`, with `X` and `Y_t` being both expected to be
+    C-contiguous, and the (equivalent) matrix product being really computed being
+    between `X` and the transpose of `Y_t`. Hence the choice of using the variable name
+    `Y_t_n_rows` rather than `Y_n_cols`.
+
+    If `out_fused_elementwise_fn` is not None, it will be applied element-wise once to
+    each element of the output array right before it is returned. This function is
+    compiled and fused into the first kernel as a device function with the help of
+    `dpex.func`. This comes with limitations as explained in:
+
+    https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html # noqa
+
+    It is expected to take one scalar argument and return one scalar value.
+
+    Likewise, if `multiply_fn` is not None it will be used in place of the scalar
+    multiply operator . It is expected to take two scalar argument and return one scalar
+    value.
+
+    `sklearn_numba_dpex.common._utils` exposes some pre-defined functions suitable to
+    be passed as `fused_elementwise_func` or `out_fused_elementwise_fn`. Passing lambda
+    functions is advised against since their compilation might not be cached.
+    """
+
+    zero = dtype(0.0)
+
+    input_work_group_size = work_group_size
+    work_group_size = _check_max_work_group_size(
+        work_group_size,
+        device,
+        required_local_memory_per_item=2 * np.dtype(dtype).itemsize,
+    )
+
+    square_window_side = math.sqrt(work_group_size)
+    square_window_side = sub_group_size * int(square_window_side / sub_group_size)
+
+    if work_group_size != input_work_group_size:
+        work_group_size = square_window_side * square_window_side
+
+    elif work_group_size != (square_window_side * square_window_side):
+        raise ValueError(
+            "Expected work_group_size to be a square of a multiple of sub_group_size "
+            f" but got sub_group_size={sub_group_size} and "
+            f"work_group_size={work_group_size}"
+        )
+
+    if multiply_fn is None:
+
+        @dpex.func
+        def multiply_fn(x, y):
+            return x * y
+
+    else:
+        multiply_fn = dpex.func(multiply_fn)
+
+    if out_fused_elementwise_fn is None:
+
+        @dpex.func
+        def out_fused_elementwise_fn(x):
+            return x
+
+    else:
+        out_fused_elementwise_fn = dpex.func(out_fused_elementwise_fn)
+
+    # The following values define the geometry of the dispatch of work items.
+    # Each work group of size (square_window_side * square_window_side) is mapped to a
+    # window of shape (square_window_side, square_window_side) that compute
+    # `work_group_size` output items in the window. Windows don't overlap, are arranged
+    # in a grid, and cover all the items to compute.
+
+    # Within a window, work items cooperate to compute one value per work item. Two
+    # temporary windows of `X` and `Y_t` values, of same shape, are loaded cooperatively
+    # in shared memory and slide along the matmul dimension to accumulate partial
+    # results, until the dimension has been completely iterated on.
+    window_shape = (square_window_side, square_window_side)
+
+    # The size of the grid of windows required to exactly cover the array of result.
+    # (note that windows at the edge will have out-of-bound items)
+    grid_n_rows = math.ceil(X_n_rows / square_window_side)
+    grid_n_cols = math.ceil(Y_t_n_rows / square_window_side)
+
+    # Amount of temporary sliding windows that are needed to accumulate the partial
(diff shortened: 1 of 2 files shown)


ogrisel:
Nice, quick benchmark vs `np.dot` (backend by OpenBLAS SGEMM) on random `float32` arrays, e.g.:

```python
import numpy as np
n, m, k = 2048, 2048, 2048
dtype = np.float32

A = np.random.randn(n, k).astype(dtype)
B = np.random.randn(k, m).astype(dtype)
C = np.empty(shape=(n, m), dtype=dtype)
%timeit np.dot(A, B, out=C)
```

fcharras:
Numpy
```
70 ms ± 2.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

sklearn_numba_dpex

```
158 ms ± 442 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

We're only twice slower :grin:

ogrisel:
sycl-blas has some Python wrapper here: https://github.com/codeplaysoftware/sycl-blas/blob/master/python_generator/py_gen_blas_gemm_launcher.py

ogrisel:
Actually this doesn't look like a wrapper. I am not sure what's the purpose of this code. Maybe a code generator.

jjerphan:
Thanks for the benchmarks, @fcharras. :)

I would be in favor of reviewing and merging this PR if there's no downside: this implementation works, is efficient, and can be made even more efficient then if we want it to.

What do you think?

fcharras:
Yes IMO it can be reviewed

fcharras:
`dpnp.matmul` on GPU provides 10* speedup on numpy. And 20* speedup on the kernel in this PR. Will probably just use `dpnp.matmul` instead.

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,277 @@
...
+    compiled and fused into the first kernel as a device function with the help of
+    `dpex.func`. This comes with limitations as explained in:
+
+    https://intelpython.github.io/numba-dpex/latest/user_guides/kernel_programming_guide/device-functions.html # noqa
+
+    It is expected to take one scalar argument and return one scalar value.
+
+    Likewise, if `multiply_fn` is not None it will be used in place of the scalar
+    multiply operator . It is expected to take two scalar argument and return one scalar
+    value.
```
jjerphan, line 43:
```suggestion
    Likewise, if `multiply_fn` is not None it will be used in place of the scalar
    multiply operator. It is expected to take two scalar arguments and return one
    scalar value.
```

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,277 @@
...
+        group_first_col_idx = dpex.get_group_id(0) * square_window_side
+        result_col_idx = group_first_col_idx + local_col_idx
+
+        # Allocate shared memory for the two sliding windows on `X` and `Y_t`
+        X_sliding_window = dpex.local.array(shape=window_shape, dtype=dtype)
+        Y_t_sliding_window = dpex.local.array(shape=window_shape, dtype=dtype)
+
+        # Index of the column of the sliding window that the current work item will be
+        # responsible for loading. The "sliding" is materialize by incrementing this
+        # index by `square_window_side` at each iteration of the main loop.
```
jjerphan, line 144:
```suggestion
        # Index of the column of the sliding window that the current work item will be
        # responsible for loading. The "sliding" is materialized by incrementing this
        # index by `square_window_side` at each iteration of the main loop.
```

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,277 @@
...
+        window_col_idx = local_col_idx
+
+        # For both `X` and `Y_t`, that are assumed to be C-contiguous, in order to
+        # optimize RW memory access patterns, it's best to ensure that contiguous work
+        # items read contiguous memory slots in both shared and global memory. This is
+        # straightforward to ensure for windows on `X`, but requires the following
+        # remapped index for `Y_T`:
```
jjerphan, line 151:
```suggestion
        # remapped index for `Y_t`:
```

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,277 @@
...
+                else:
+                    for idx in range(square_window_side):
+                        output += multiply_fn(
+                            X_sliding_window[local_row_idx, idx],
+                            Y_t_sliding_window[local_col_idx, idx],
+                        )
+
+            dpex.barrier(dpex.LOCAL_MEM_FENCE)
+
+        # Write result
```
jjerphan, line 224:
```suggestion
        # Save the value of the scalar product of the result_row_idx-th row of X
        # and of the result_col_idx-th column of Y.
```

jjerphan (commented):
A few comments. Do we need to introduce this kernel since `dnpn.matmul` exist?

fcharras:
> A few comments. Do we need to introduce this kernel since `dnpn.matmul` exist?

I wonder. I don't understand how the performance look so weak here, while our kmeans seems to work well, at least it compares as one could expect with cpu implementation.

I suppose the performances will not be as bad with typical input dimension for KNN (with `n_cols` = `n_features` is much lesser than the other two dimensions), and then it might actually be better to use this kernel because we can fuse `multiply_fn` and `elementwise_fn` to compute pairwise distances in one kernel call.

fcharras:
I'm currently working on an update on this. I think it's actually possible to achieve near dpnp performance with numba_dpex.

fcharras:
Also I've opened a discussion at `numba_dpex` see https://github.com/IntelPython/numba-dpex/issues/984

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,286 @@
...
+import numba_dpex as dpex
+import numpy as np
+
+from sklearn_numba_dpex.common._utils import _check_max_work_group_size
+
+zero_idx = np.int64(0)
+
+
+@lru_cache
+def make_matmul_2d_kernel(
```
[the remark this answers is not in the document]
ogrisel: I am not sure we need the full GEMM API in most cases and specializing it for simple MM might be more efficient for the most common cases (while keeping the code simpler).

I think a full GEMM kernel (or family of kernels to be specialized for subcases) can be done upstream in numba_dpex if needed.

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,814 @@
...
+        work_item_row_idx = work_item_idx // sub_group_size
+        work_item_col_idx = work_item_idx % sub_group_size
+
+        group_idx = dpex.get_group_id(zero_idx)
+
+        # Get indices of the row and the column of the top-left corner of the sub-array
+        # of results covered by this work group
+        group_row_idx = group_idx // global_grid_n_cols
+        group_col_idx = group_idx % global_grid_n_cols
+        # TODO: improve cache locality ?
```
ogrisel, line 344: nit

```suggestion
        # TODO: improve cache locality?
```

sklearn_numba_dpex/common/matmul.py
```diff
@@ -0,0 +1,814 @@
...
+    # `arithmetic_intensity_multiplier_X + arithmetic_intensity_multiplier_Y`
+    # more items in the work group shared memory, per item in this work group.
+
+    # Those values should be set high enough so that the kernel is compute-bound
+    # rather than memory-bound.
+
+    # Setting it too high might pressure local or private memory too much for it to be
+    # worth.
+    arithmetic_intensity_multiplier_X = 4
+    arithmetic_intensity_multiplier_Y = 2
```
ogrisel, line 112: Let's make those kwargs of `make_matmul_2d_kernel` with the same default values so that it will be possible in the future to do run some kind of auto-tuning grid search.

ogrisel (approved):
I won't review the core of the kernel itself but +1 for merge with stronger non-regression tests as explained below (at least a few more test cases with larger shapes and maybe non-integer, positive and negative input values).

sklearn_numba_dpex/common/tests/test_matmul.py
```diff
@@ -0,0 +1,109 @@
...
+        ((5, 8), (8, 4)),
+        ((5, 10), (10, 4)),
+    ],
+)
+@pytest.mark.parametrize("dtype", float_dtype_params)
+def test_matmul_2d(test_input_shapes, work_group_size, sub_group_size, dtype):
+    X_shape, Y_shape = test_input_shapes
+
+    X = _arange_reshaped(X_shape, dtype)
+    Y = _arange_reshaped(Y_shape, dtype)
```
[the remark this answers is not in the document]
fcharras: It doesn't cost much more to have both the `_arange_reshaped` and `random` tests so I can add both. (I think `_arange_reshaped` is enough, but above that it is handier for debug than random values so I'd like to keep it along)

I think the boundary effects are already all covered by this set of test cases (the inputs are smalls but the tests also use small values for `work_group_size` / `sub_group_size`)

fcharras:
Latest push should have addressed comments, this version is mergeable IMO.

- [x] improved tests
- [x] removed WIP work on cache locality (and pushed it in the matmul_kernel_wip on this repo)
- [x] exposed all performance parameters

ogrisel (approved):

PR #104, opened 2023-03-02
FIX fix performance regression seen in `numba_dpex==0.20.0dev3`

fcharras:
Back to monkey patching :grinning: 

Contrary to previous patches for `0.19.0`, I can't think of an easy test to check that the patch works or to notify when the patch can be reverted.

I checked that this solves performance issue on edge cloud and the benchmark is still fine locally. (but it doesn't seem to fix the slight added overhead that we've observed locally, but it's not much of an issue),


ogrisel (approved):
Assuming the test pass, LGTM since this hack is only meant to be temporary.

PR #107, opened 2023-04-20
Improve performance accross several kernels with better parameters

fcharras:
With this PR:

- <del>our matmul achieves 90% dpnp performance on iGPU</del> See: https://github.com/soda-inria/sklearn-numba-dpex/pull/107#issuecomment-1519389111.
- catastrophic performances on max series gpu is fixed of KMeans

also fix compatibility with latest versions of `dpctl`.

Closes https://github.com/IntelPython/numba-dpex/issues/984

### DIFF
--- a/sklearn_numba_dpex/common/matmul.py
+++ b/sklearn_numba_dpex/common/matmul.py
@@ -106,7 +106,7 @@ def make_matmul_2d_kernel(
     - https://triton-lang.org/master/getting-started/tutorials/03-matrix-multiplication.html  # noqa
 
     """
-    if work_group_size in [None, "max"]:
+    if work_group_size is None:
         work_group_size = 512
     if sub_group_size is None:
         sub_group_size = 32
--- a/sklearn_numba_dpex/common/matmul.py
+++ b/sklearn_numba_dpex/common/matmul.py
@@ -36,8 +36,8 @@ def make_matmul_2d_kernel(
     # rather than memory-bound.
     # Setting it too high might pressure local or private memory too much for it to be
     # worth.
-    arithmetic_intensity_multiplier_X=2,
-    arithmetic_intensity_multiplier_Y=4,
+    arithmetic_intensity_multiplier_X=1,
+    arithmetic_intensity_multiplier_Y=1,
     # Width of the window of values of Y_t stored in registries - lower means less
     # values in registries, but more iterations in main loop.
     private_Y_t_sliding_window_width=1,  # must divide `sub_group_size`,
@@ -107,9 +107,9 @@ def make_matmul_2d_kernel(
 
     """
     if work_group_size is None:
-        work_group_size = 512
+        work_group_size = 128
     if sub_group_size is None:
-        sub_group_size = 32
+        sub_group_size = 4
 
     # NB: The following implementation not only works for the best parameters found so
     # far but for other combinations too, to enable exhaustive grid searches on all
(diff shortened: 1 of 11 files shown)


sklearn_numba_dpex/common/matmul.py
```diff
@@ -195,18 +195,8 @@ def out_fused_elementwise_fn(x):
...
-    )
-    local_Y_t_sliding_window_shape = (
-        result_window_width,
-        2 * sub_group_size,  # allocate twice the space to let padding against bank
-        # conflicts
-    )
-    # TODO: it's surprising to see bank conflicts here, however the padding add
-    # 10% better performance on iGPUs. Why ?
+    local_X_sliding_window_shape = (sub_group_size, result_window_height)
+    local_Y_t_sliding_window_shape = (sub_group_size, result_window_width)
```
fcharras, line 199: NB: this PR also transposes the layout of the sliding windows. I don't understand why, but it seems to be half the reason we've unlocked better performance. :shrug:

sklearn_numba_dpex/common/random.py
```diff
@@ -33,7 +33,7 @@ def get_random_raw(states):
 
     Note, this always uses and updates state states[0].
     """
-    result = dpt.empty(sh=(1,), dtype=np.uint64, device=states.device)
```
fcharras, line 36: This prepares compatibility with latest versions of `dpctl`.

Straight removing `sh=` rather than replacing with `shape=` enables compatibility with all version before and after the breaking changes.

ogrisel:
> catastrophic performances on max series gpu is fixed

This is great! How does it compare with dpnp/MKL on max series now?

sklearn_numba_dpex/common/random.py
ogrisel: Note that in this case, passing 1 instead of `(1,)` might also work (I think it does for numpy).

ogrisel (approved):
LGTM!

fcharras:
Sorry, I meant

> catastrophic performance of kmeans

so nothing new here :-( I couldn't check matmul performance on max series yet with the new parameters or perform another autotune.

On Flex it's not as good as 90% of dpnp performance, it's about 50% performance with those parameters, but it still looks decent.

It's likely that to aim at near-maximum performance on all devices, we'd need to keep a dictionary of best grid-searched parameters for each device, or let the user perform an auto-tune and cache the results in a local config folder before using the software. (like it seems triton does already)

sklearn_numba_dpex/kmeans/kernels/lloyd_single_step.py
```diff
@@ -111,23 +111,10 @@ def make_lloyd_single_step_fixed_window_kernel(
...
     # NB: for more details about the privatization strategy the following variables
     # refer too, please read the inline commenting that address it in the kernel
     # definition.
 
-    # Ensure that the memory allocated for privatization will not saturate the global
-    # memory cache size. For this purpose we limit the number of private copies to
-    # a fraction of the available global memory cache size.
-    n_centroids_private_copies = (
-        global_mem_cache_size * centroids_private_copies_max_cache_occupancy
-    ) // n_cluster_bytes
```
fcharras, line 129: Experiments have shown that having a too low number of private copies results in catastrophic performance. Compared to that, saturating memory caches is not much of a worry.

fcharras:
The good performance announcement of the `matmul` was a false flag, there were a bug for some combinations of parameters that made it look fast, but the output was wrong.

I've updated all parameters for `KMeans`, `matmul` and `topk` with best parameter found for the `PVC` gpu. I've found best parameters found on the `PVC` to translate reasonably well to other GPUs, while the opposite is not true, optimizing on iGPU or Flex can give catastrophic performance on PVC.

The `matmul` remains consistently at 25-50%performance of `dpnp`.

ogrisel (approved):
Don't have time to test it on my iGPU right now but LGTM!

Issue #94, opened 2023-02-22
Leverage the lightweight dpcpp-llvm-spirv dependency


ogrisel:
https://github.com/IntelPython/dpcpp-llvm-spirv

should be available from the `dppy/label/dev` conda channel.

It should be much lighter weight than the `"intel::dpcpp_linux-64"` we currently use as a dependency for `sklearn_numba_dpex`.

PR #109, opened 2023-04-24
Fix docker build pipeline

fcharras:
Numba 0.57 compatibility issues are already reported upstream see https://github.com/IntelPython/numba-dpex/issues/1000


docker/Dockerfile
```diff
@@ -60,7 +60,7 @@ ARG INTEL_SCIPY_VERSION="==1.7.3"
 # content of the intel pypi or not. (this Dockerfile currently assumes that
 # it is not limited to intel pypi. The opposite assumption seems to create
 # impossible build conditions.)
-ARG INTEL_NUMBA_VERSION=""
+ARG INTEL_NUMBA_VERSION="<0.57"
```
ogrisel, line 63: Please link to an issue that explains why it's needed:

- if it's a known issue upstream, just linked to that,
- if it's something we need to updated in our code, maybe open an issue with the traceback one would typically observe if we remove the pinning.

ogrisel (approved):

docker/Dockerfile
fcharras: See issue OP :grin:

fcharras:
I'm still not building it locally

There's something wrong with dpnp build that requires further debug

fcharras:
Latest commits both address latest issues and bump to latest releases.

I'd like to bump aggressively to latest commits of master branch for `numba_dpex` so it includes cache performance fixes, will revert to a fixed tag once it's officially released.

fcharras:
Fixes https://github.com/soda-inria/sklearn-numba-dpex/issues/94

fcharras:
Not yet functional, see https://github.com/IntelPython/dpctl/issues/1190

However the conda install works.

fcharras:
Or if the the `intel-opencl-rt` issue is not fixed upstream it might be the opportunity to try the official [`numba_dpex` docker image ](https://intelpython.github.io/numba-dpex/latest/user_guides/docker.html) but I'm afraid we would loose some flexibility.

fcharras:
Installers work again with 2023.2.0 release, but there's a regression in latest `numba_dpex` release that prevent some kernels to run on CPU (e.g. our lloyd fused kernel), see https://github.com/IntelPython/numba-dpex/issues/1068

fcharras:
Things work on my local environment, I'll merge so that the automated build pipeline can resume, and the conda install instruction are fixed.

PR #110, opened 2023-04-26
ENH Reuse TopK buffers

fcharras:


sklearn_numba_dpex/common/kernels.py
```diff
@@ -36,25 +36,41 @@ def elementwise_ops(data):
...
+    def fill(data):
         data = dpt.reshape(data, (-1,))
-        initialize_to_zeros_kernel[global_size, work_group_size](data)
+        fill_kernel[global_size, work_group_size](data)
+
+    return fill
+
+
+@lru_cache
+def make_range_kernel(n_items, work_group_size):
```
ogrisel, line 61: Maybe `range` => `arange` to match the numpy function name?

Otherwise one can get confused with the "range" name from SYCL terminology.

ogrisel (approved):
The docstring for the `_make_get_topk_kernel` kernel looks good but as far as I can see, you did not add a unit test that checks that the closure tricks works as expected when doing repeated calls to the same closure, in particular for both possible values of `reuse_result_buffer`.

Other than that, LGTM (I just did a very shallow review).

fcharras:
The unit test is a bit technical with the closure-style definition of things everywhere but it should be possible, I'll look into it.

PR #118, opened 2023-10-05
MNT Update dependencies

fcharras:
Update Dockerfile, patch around inline optimization issues, and hack around drop of atomics on CPU to keep a working CI (by pinning CI on an old working version).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,7 @@
 #
 # TODO: make a custom build with other base images to test compatibility
 # with other OSes.
-ARG BASE_IMAGE_VERSION=latest-19f98074bfb71824c8fec117f2f5e6fd1d2d3da9
+ARG BASE_IMAGE_VERSION=latest-57b016e6e031950e5e5abda2d69afc355e7c7b61
 ARG BASE_IMAGE_TAG=ghcr.io/intel/llvm/ubuntu2204_intel_drivers
 
 ARG BASE=${BASE_IMAGE_TAG}:${BASE_IMAGE_VERSION}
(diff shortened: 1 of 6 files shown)


docker/Dockerfile
```diff
@@ -62,35 +62,34 @@ ARG ONEAPI_INSTALL_DIR=/opt/intel/oneapi
...
 ARG DPCPP_LLVM_SPIRV_GIT_URL=https://github.com/IntelPython/dpcpp-llvm-spirv/
 
-
 # Version of other python packages explicitly installed either within the
 # build environment or the runtime environment.
 
 ARG NUMPY_VERSION="<1.25"
 ARG NUMBA_VERSION=""
-ARG CYTHON_VERSION="<3"
+ARG CYTHON_VERSION=""
```
ogrisel, line 92: If Cython is used to build scikit-learn, note that there are a few performance regressions with Cython 3. So we might want to stick to the latest Cython 2 for the time being.

ogrisel (approved):
If it works, LGTM :)

ogrisel:
Apparently it doesn't work:

```
WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
WARNING: Requirement '/opt/numba_dpex/cpu-stable/numba_dpex*.whl' looks like a filename, but the file does not exist
ERROR: numba_dpex*.whl is not a valid wheel filename.
```

fcharras:
It's normal that the test don't pass yet, we must merge first so that the runner build and update the image to our repo, and re-run the pipelines after that. I'll perform a few final tests locally before.

docker/Dockerfile
fcharras: cython is not installed in the final image, so it's not related to the cython one can use to build sklearn later on in a container. Most recent dpctl has merged work that makes it compatible with cython>=3 so I think we can keep that.

PR #120, opened 2023-10-06
MNT remove hacks after some stability issues have been fixed upstream

fcharras:
However, add NUMBA_DPEX_OPT=0 when running tests on CPU


ogrisel (approved):
LGTM if CI is green.

